PR #3, opened 2014-08-18
Cssshapes

devnook:
Adding a CSS Shapes level 1 sample.


jeffposnick:
This is what I'm seeing for the last two examples (using `content-box` and `padding-box`) using Chrome 38. Just wanted to make sure it's intended.

image

css-shapes/index.html
```diff
@@ -0,0 +1,248 @@
...
+        shape-outside: polygon(0 0, 100% 0, 100% 100%, 0 100%) content-box;
+      }
+    </style>
+    <div class="shape box content-box">
+      <p>content</p>
+    </div>
+    <p>
+      This shape uses <code>content-box</code>, defined in the
+      <code>shape-outside</code> property. The content is in red,
+      content in yellow and margin area in light blue.
```
jeffposnick: Should this say "_padding in yellow_" instead of "_content in yellow_"?

devnook: Fixed. as for the flowing text, it is there to show different spacing of the element, so it's intended.

devnook:
PTAL

PR #5, opened 2014-08-19
Encoding API Sample

jeffposnick:
Sample code for the [Encoding API](http://www.chromestatus.com/feature/5714368087982080), available in Chrome 38+.

A live demo is currently available at https://jeffposnick.github.io/samples/encoding-api/index.html


jeffposnick:
Actually, I need to rethink my branch setup... will refile this later.

PR #10, opened 2014-08-20
Added back in butterfly as part of WebP+picture demo

pearlchen-old:
There are a few use cases for `<picture>` so I added in a second example for file format support. 

Is it okay to have multiple `<!-- // [START code-block] -->` blocks sprinkled throughout the text? Added `<section>` tags with `id` for URL anchor support.


picture-element/index.html
```diff
@@ -13,37 +13,56 @@
...
-      <source media="(min-width: 650px)" srcset="images/kitten-large.png">
-      <source media="(min-width: 465px)" srcset="images/kitten-medium.png">
-      <!-- img tag for browsers that do not support picture element -->
-      <img src="images/kitten-small.png" alt="a cute kitten">
-    </picture>
-    <!-- // [END code-block] -->
+      <p>In the second example below, the browser will display a WebP image 
+      instead of a JPG when the browser supports WebP.</p>
 
+      <!-- // [START code-block] -->
```
jeffposnick: The whole extracting code snippets via comment tags thing isn't implemented yet, but I think that using distinct identifiers for each distinct snippet will end up making the most sense. So `code-block-responsive-sizes` and `code-block-fallback-formats` or something along those lines.

pearlchen-old: Sounds good. I updated to:
`<!-- // [START code-block-source-media] -->`
and
`<!-- // [START code-block-source-type] -->`

PR #11, opened 2014-08-20
Symbols (ES6) sample, simulating private variables in a class.

jeffposnick:
@devnook, @addyosmani, & co., here's my attempt on some sample code related to ES6 Symbols.

Doing something related to private member variables seemed relevant, so I went that route. Definitely interested in feedback about the implementation, though—I would not be surprised if it could be improved.

Live demo at https://rawgit.com/jeffposnick/samples/symbols-es6/symbols-es6/index.html


jeffposnick:
After some discussion it's been determined that this is not a use case we want to encourage for Symbols, so I'm abandoning this PR.

PR #12, opened 2014-08-21
Map, Set, and for...of iterator samples.

jeffposnick:
@addyosmani, if you have any ideas as to someone who might be able to look this over for best practices, it would be appreciated!


addyosmani:
Have these gone through code-review yet? If not, it might be useful for us to keep future PRs open until we've had a chance to verify the source in the sample can't be improved further.

Otherwise, great work on shipping these so quickly.

jeffposnick:
They were reviewed off-GitHub by some folks with experience working on the ES6 spec, and the recent updates took their feedback into account.

I thought I mentioned that in the comment that accompanied the merge, but I can't find that comment displayed in the GitHub UI (which is strange).

PR #13, opened 2014-08-21
Added example for pixelated image rendering.

pearlchen-old:
For feature: http://www.chromestatus.com/feature/5118058116939776

Preview: http://pearlchen.github.io/samples/image-rendering-pixelated/

### DIFF
--- a/image-rendering-pixelated/README.md
+++ b/image-rendering-pixelated/README.md
@@ -1,0 +1,6 @@
+Pixelated Image Rendering Sample
+===
+
+See https://googlechrome.github.io/samples/image-rendering-pixelated/index.html for a live demo.
+
+Learn more at http://www.chromestatus.com/feature/5118058116939776
--- a/image-rendering-pixelated/index.html
+++ b/image-rendering-pixelated/index.html
@@ -1,0 +1,44 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>Pixelated Image Rendering Sample</title>
+
+    <link href='//fonts.googleapis.com/css?family=Roboto+Condensed:300italic,400italic,700italic,400,300,700' rel='stylesheet' type='text/css'>
+    <link rel="stylesheet" href="../styles/main.css">
+  </head>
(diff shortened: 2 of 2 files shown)


any-pointer-any-hover/index.html
```diff
@@ -0,0 +1,79 @@
...
+    <title>'any-pointer' and 'any-hover' Media Queries Sample</title>
+
+    <link href='//fonts.googleapis.com/css?family=Roboto+Condensed:300italic,400italic,700italic,400,300,700' rel='stylesheet' type='text/css'>
+    <link rel="stylesheet" href="../styles/main.css">
+  </head>
+  <body>
+    <h1>'any-pointer' and 'any-hover' Media Queries Sample</h1>
+    <p>Available in <a href="http://www.chromestatus.com/feature/6460705494532096">Chrome 38+</a></p>
+
+    <p>The <code>any-pointer</code> and <code>any-hover</code> give web authors the ability to feature detect </p>
```
jeffposnick: This looks like an incomplete sentence.

any-pointer-any-hover/index.html
```diff
@@ -0,0 +1,79 @@
...
+
+      <dt>hover</dt>
+      <dd></dd>
+
+    </dl>
+
+    <!-- // [START code-block] -->
+    <style>
+
+      /* Make radio buttons and check boxes larger if we have an inaccurate pointing device */
```
jeffposnick: Could you add some `<input type="checkbox">` and `<div class="menu>` elements to the page to illustrate how the styles influence the way the elements are rendered?

pearlchen-old:
@jeffposnick Sorry, I didn't mean to actually submit a PR, just an issue. Here's my question from the spreadsheet:

"Here's a draft sample code. I feel like it'll be more appropriate to break them into 2 separate samples/files though. Logistically, what do you think the best approach is? 2 separate directories? Or keep in the same place?"

jeffposnick:
A separate directory for each distinct feature fits with the overall model we have going right now.

If you could split out the pixelated demo (which I think is good to go) from the any-pointer/hover demo (which probably still needs work) and just include the pixelated demo in this PR that would be ideal. Creating separate git branches for each sample should make that easier.

pearlchen-old:
Oh, I see what happened. I pushed the wrong local branch to my gh-pages. It's back to just the pixelated image rendering example.

PR #14, opened 2014-08-27
Brings the placeholder index.html more in line with WSK's basic.html

### LINKED ISSUE #4: Use Web Starter Kit as the placeholder baseline.
At minimum I would suggest we go for the minimal options in https://github.com/google/web-starter-kit/blob/master/app/basic.html, if not the full layout options in https://github.com/google/web-starter-kit/blob/master/app/index.html.

basic.html support would just mean a few minor new additions atm. cc @PaulKinlan

jeffposnick:
@addyosmani & @PaulKinlan, this updates the placeholder `index.html` to match what's found in https://github.com/google/web-starter-kit/blob/master/app/basic.html

If this looks to be what you had in mind, I'll update the individual samples to adopt the changes. 

Closes #4


placeholder-name/index.html
```diff
@@ -1,34 +1,60 @@
...
+
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-status-bar-style" content="black">
+    <link rel="apple-touch-icon-precomposed" href="../images/apple-touch-icon-precomposed.png">
+
+    <!-- Tile icon for Win8 (144x144 + tile color) -->
+    <meta name="msapplication-TileImage" content="images/touch/ms-touch-icon-144x144-precomposed.png">
+    <meta name="msapplication-TileColor" content="#3372DF">
+
+    <link rel="icon" href="../images/favicon.ico">
```
jeffposnick, line 46: I'm explicitly including this `<link>` to `favicon.ico` since it's not in the browser default `/favicon.ico` location.

I'm not sure if there's a specific order that I need to put this in in relation to the other various icon definitions, or whether it's strictly needed, but I'm assuming if there are some browsers that don't recognize the `sizes="192x192"` icon then this would be used.

Happy to hear if I'm wrong and this isn't needed.

addyosmani:
LGTM. Thanks for putting this together!

PR #17, opened 2014-08-29
Adding screen orientation sample

devnook:


screen-orientation/index.html
```diff
@@ -0,0 +1,86 @@
...
+
+    <script>
+      var lock = function() {
+        /* // [START lock-orientation] */
+        screen.orientation.lock('landscape').then(
+          function() {
+            alert('Orientation locked!');
+          },
+          function(err) {
+            alert(err.message);
```
jeffposnick, line 70: The Promise always rejects with a "This page needs to be fullscreen in order to lock orientation." error. (Tested on Chrome 38.0.2125.57 on Android.)

I think that's part of the spec. If you wanted to demonstrate the fullscreen + orientation lock process, I'm sure that would be useful. Alternatively, you could just remove the code that attempt to lock and add some explanatory text about needing to go fullscreen first.

devnook, line 70: Done

devnook:
PTAL

jeffposnick:
Looks great now. Thanks! :+1:

PR #18, opened 2014-08-29
Display an error message when the TextDecoder interface isn't available.

jeffposnick:
My intention for this sample was to show the native functionality on supported browsers, so I'm not including the [polyfill](https://github.com/inexorabletash/text-encoding), and just display a message if the browser doesn't support `TextDecoder`.

@ebidel let me know if you think these samples **should** demonstrate use of polyfills (when available).

### DIFF
--- a/encoding-api/index.html
+++ b/encoding-api/index.html
@@ -30,16 +30,20 @@
     <pre id="results"></pre>
 
     <script>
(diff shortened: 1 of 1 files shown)


ebidel:
LGTM. No, I don't think we need to demonstrate use of a polyfill. If one is available, it's good to link off to it but the purpose of these samples is to show off the API for the first time!

PR #19, opened 2014-09-04
Service Worker registration sample

jeffposnick:
@slightlyoff @jakearchibald @KenjiBaheux

Here's a catch-all sample to illustrate some Service Worker lifecycle considerations—at least as well as I understand them, and based on how they're implemented in the current Chrome Canary build. Some aspects could be broken out into smaller pieces, but I wanted to solicit overall feedback and make sure I wasn't completely misunderstanding the bigger picture before doing that.

This sample doesn't attempt to cover anything related to caching or the Push API.

There are some inline PR comments on lines that I am particularly unsure about.

You can view a live demo at https://jeffposnick.github.io/samples/service-worker/registration/

### DIFF
--- a/service-worker/registration/index.html
+++ b/service-worker/registration/index.html
@@ -1,0 +1,190 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>Service Worker Sample: Registration</title>
+
+    <link rel="stylesheet" href="../../styles/main.css">
+    <style>
+      .registration {
+        margin: 0.5em;
+        padding: 0.5em;
+        border-radius: 0.5em;
+        background-color: rgb(235, 235, 235);
+      }
+
+      .installing {
+        background-color: rgb(225, 225, 225);
+      }
+
+      .installed {
+        background-color: rgb(215, 215, 215);
+      }
+
+      .activating {
+        background-color: rgb(205, 205, 205);
+      }
+
+      .activated {
+        background-color: rgb(125, 215, 125);
+      }
+
+      .redundant {
+        background-color: rgb(215, 125, 125);
+      }
+    </style>
+  </head>
+  <body>
+    <script src="http://www.polymer-project.org/platform.js"></script>
+    <link rel="import" href="http://www.polymer-project.org/components/polymer/polymer.html">
+
+    <h1>Service Worker Sample: Registration</h1>
+    <p>Available in <a href="http://www.chromestatus.com/feature/6561526227927040">Chrome 39+</a></p>
+
+    <!-- // [START code-block] -->
+    <template id="page-template" is="auto-binding">
+      <template if="{{ error }}">
+        <h3>Error: {{error}}</h3>
+      </template>
+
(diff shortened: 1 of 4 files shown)


service-worker/registration/index.html
```diff
@@ -0,0 +1,190 @@
...
+      };
+
+      if ('serviceWorker' in navigator) {
+        Object.keys(scriptToScopes).forEach(function(script) {
+          var scopes = scriptToScopes[script];
+
+          scopes.forEach(function(scope) {
+            navigator.serviceWorker.register(script, { scope: scope }).then(
+              function(registration) {
+                [registration.installing, registration.waiting, registration.active].forEach(function(serviceWorker) {
```
jeffposnick, line 149: I have a feeling that I shouldn't be paying attention to all three of `.installing`, `.waiting`, and `.active`, but I'm not sure which ones will correspond to a `ServiceWorker` that I should actually care about. (It's not always going to be `.installing`, is it?)

service-worker/registration/service-worker-one.js
```diff
@@ -0,0 +1,55 @@
...
+
+  switch (e.data) {
+    case 'update':
+      // Not yet implemented in Chrome Canary.
+      // self.update();
+      e.ports[0].postMessage('self.update() would have been called on the Service Worker (not currently implemented).');
+    break;
+
+    case 'getAll':
+      self.clients.getAll().then(function(serviceWorkerClients) {
```
jeffposnick, line 32: I think @KenjiBaheux mentioned that `getAll()` should be working, but I'm not seeing it return any `ServiceWorkerClient`s when testing this out.

service-worker/registration/index.html
```diff
@@ -0,0 +1,190 @@
...
+
+      template.sendMessage = function(e) {
+        var model = e.target.templateInstance.model;
+        var message = template.messageValues[model.i];
+        template.messageValues[model.i] = '';
+
+        postMessageToWorker(message, model);
+      };
+
+      var scriptToScopes = {
```
jeffposnick, line 137: Thinking a bit more about what I initially wrote, is registering multiple of script/scope combinations overly contrived? If it's not a scenario that would come up much in the real world, I could retool this sample so that it focuses on a single registration, forces an update, and then shows how the updated `ServiceWorker` transitions from waiting to active.

Would that make for a more meaningful sample?

jeffposnick:
Okay, I've spent another day wrapping my head around the recent `ServiceWorker` changes, especially around registration and the update flow, and I get the feeling that this sample needs some changes to demonstrate anything that would be useful in the real world.

I'm going to take another stab at a revised PR. I'll leave this open for now in case anyone is curious, but will likely abandon it next week.

service-worker/registration/service-worker-one.js
KenjiBaheux, line 32: clients is always empty which is unexpected (afaik)
Filed crbug.com/412338

service-worker/registration/index.html
KenjiBaheux, line 137: Yes, I think we should focus on the simplest possible use case. Other demos might address the more complex cases.

service-worker/registration/index.html
KenjiBaheux, line 149: I could be wrong but eventually what really matters in the simplest case is "activated". This would be the point from which the service worker is effectively in control.

PR #20, opened 2014-09-10
Basic Registration Sample

jeffposnick:
@jakearchibald @KenjiBaheux

The first of the micro-samples, which will likely be linked to from [Know Your Service Worker](https://jeffposnick.github.io/know-your-service-worker/cookbook.html) but are being kept in this repo for consistency with the overall samples strategy.

Other micro-samples to follow, roughly following what's outlined in [this list](https://jeffposnick.github.io/know-your-service-worker/cookbook.html).

You can view a live version of these changes at https://rawgit.com/jeffposnick/samples/sw-registration-basic/service-worker/registration/index.html

### DIFF
--- a/service-worker/registration/README.md
+++ b/service-worker/registration/README.md
@@ -1,0 +1,5 @@
+Service Worker Sample: Basic Registration
+===
+See https://googlechrome.github.io/samples/service-worker/registration/index.html for a live demo.
+
+Learn more at http://www.chromestatus.com/feature/6561526227927040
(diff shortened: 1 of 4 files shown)


jeffposnick:
I'd like to move ahead on the other samples, so I'm going to be merging this in. There's another PR incoming with similar content, so feel free to comment on that one if there's something fundamental that needs to change, and I'll apply those changes to all of the samples.

PR #21, opened 2014-09-15
Detailed Service Worker registration example.

jeffposnick:
@jakearchibald @KenjiBaheux

This sample takes things a step further by attempting to illustrate aspects of a Service Worker registration:
- Whether SW is available
- Whether the current page is already controlled by a SW
- Whether the current registration attempt represents the `installing`, `waiting`, or `active` SW
- The state transitions of the SW associated with the current registration

You can view a live demo at https://rawgit.com/jeffposnick/samples/sw-registration-events/service-worker/registration-events/index.html


service-worker/registration-events/index.html
```diff
@@ -0,0 +1,163 @@
...
+    navigator.serviceWorker.register('service-worker.js', { scope: './' }).then(
+      function(registration) {
+        // At this point, registration has taken place.
+        // The Service Worker will not handle requests until this page and any
+        // other instances of this page (in other tabs, etc.) have been
+        // closed/reloaded.
+        document.querySelector('#register').textContent = 'succeeded';
+
+        var serviceWorker;
+        if (registration.installing) {
```
jeffposnick: Is this the correct way to get the SW associated with the current registration? Is it safe to assume that the current registration corresponds to exactly one of `.installing`, `.waiting.`, or `.active`, in that order?

service-worker/registration-events/index.html
```diff
@@ -0,0 +1,168 @@
...
+
+  <link rel="icon" href="../../images/favicon.ico">
+
+  <link rel="stylesheet" href="../../styles/main.css">
+</head>
+
+<body>
+<h1>Service Worker Sample: Detailed Registration</h1>
+
+<p>Available in <a href="http://www.chromestatus.com/feature/6561526227927040">Chrome 39+</a></p>
```
KenjiBaheux: @jeffposnick 
"Available" might be confusing, perhaps change it to:
 Available behind the experimental web platform features flag in Chrome 39+ (<a href="http://www.chromium.org/blink/serviceworker/getting-started">learn more</a>).

service-worker/registration-events/index.html
```diff
@@ -0,0 +1,168 @@
...
+
+<!-- // [START code-block] -->
+<div class="output">
+  <h2>Service Worker Availability</h2>
+  <p>
+    Service Workers <strong id="availability"></strong> available in your browser.
+    <em>(Whether <code>'serviceWorker' in navigator</code>.)</em>
+  </p>
+
+  <h2>Is This Page Controlled?</h2>
```
KenjiBaheux: Perhaps, make this more explicit? =>
`<h2>Is This Page Controlled by a Service Worker?</h2>`

service-worker/registration-events/index.html
```diff
@@ -0,0 +1,168 @@
...
+
+<body>
+<h1>Service Worker Sample: Detailed Registration</h1>
+
+<p>Available in <a href="http://www.chromestatus.com/feature/6561526227927040">Chrome 39+</a></p>
+
+<p>
+  This sample provides some insight into the events involved in a typical Service Worker registration.
+  The script <code>service-worker.js</code> is registered to handle the URL scope <code>./</code>.
+  It is "safe" to register the same script multiple times, so it's not necessary to check for a
```
KenjiBaheux: @jeffposnick 
navigator.serviceWorker.getRegistration has been implemented recently in Chrome canary (39).

It's still true that it is safe to call register with the same script and scope multiple times but should we advertise the more natural way of doing things instead (i.e. getRegistration, register if undefined)?

The advantage of checking with getRegistration first is that if there are no Service worker registered then you don't inadvertently register one prematurely: you can do other things before eventually deciding to register one.

/cc @jakearchibald

service-worker/registration-events/index.html
```diff
@@ -0,0 +1,168 @@
...
+  </p>
+  <p>
+    The current registration represents the <strong id="kind"></strong> Service Worker.
+    <em>
+      (Whether <code>registration.installing</code>, <code>registration.waiting</code>,
+      or <code>registration.active</code> is set.)
+    </em>
+  </p>
+  <p>
+    <em>Tip:</em> Try reloading this page to activate an installing/waiting Service Worker and
```
KenjiBaheux: @jeffposnick 
This is a bit tricky. See Jake's [blog post](http://jakearchibald.com/2014/using-serviceworker-today/) "in practice section".

To be safe, I think that we should rewrite this:
- With a newly registered Service Worker in the activated state, try reloading the page to have it start controlling the page. You can also reload ...

And perhaps add the following note:
- With a updated Service Worker in the installed state, try closing and re-opening the tab to have the newer Service Worker start controlling the page. Alternatively, use shift + reload followed by reload.

cc/ @jakearchibald

service-worker/registration-events/index.html
KenjiBaheux: This might need to be changed slightly if we go for getRegistration first then register if needed.

If I'm not mistaken, depending on the circumstances, there can be one of each (one .installing, one .waiting, one active). For instance, when you register again with an updated script, you'll have installing (new SW) and active (old SW) and then waiting (new SW) and active (old SW).

That being said, checking for installing, waiting and active in that order should give you the most recent SW.

cc/ @jakearchibald : I might be a bit fuzzy here, help appreciated.

service-worker/registration-events/service-worker.js
```diff
@@ -0,0 +1,20 @@
+// In Chrome, you can view these log messages and many more useful pieces of
+// debugging info at chrome://serviceworker-internals
+
+self.addEventListener('install', function(e) {
+  // This event will be fired once when this version of the script is first registered for
+  // a given URL scope.
+  // It's an opportunity to initialize caches and prefetch data, if desired.
+  console.log('Install event:', e);
```
KenjiBaheux, line 12: We should call out the fact that you shouldn't do anything that may disrupt an existing Service Worker already controlling the page.

Should we introduce waitUntil here?

cc/ @jakearchibald

service-worker/registration-events/service-worker.js
```diff
@@ -0,0 +1,20 @@
...
+});
+
+self.addEventListener('activate', function(e) {
+  // This event will be fired once when this version of the script is first registered for
+  // a given URL scope.
+  // It's an opportunity to clean up any stale data that might be left behind in self.caches
+  // by an older version of this script.
```
KenjiBaheux, line 19: From Jake's blog post, we should call out the fact that ops here might delay the page load and resources. So, the idea is to do as much as possible during the oninstall phase where there isn't such issue.

KenjiBaheux:
@jeffposnick Sorry for the late reply, I made a few comments.

jeffposnick:
Thanks for the detailed comments, @KenjiBaheux. Since Chrome's implementation is changing so quickly, the "late reply" just means that I have a chance to sync up on the most recent changes (`getRegistration()` going live, etc.).

I'll push out some revisions.

service-worker/registration-events/index.html
jeffposnick: I'm going to leave as-is, as I haven't seen anything in the public issue tracker discussion indicating that `getRegistration()` followed by a conditional `register()` is the One True Pattern.

Happy to switch if I've missed the latest recommendation.

service-worker/registration-events/index.html
jeffposnick: I've switched it to Chrome 40+.

service-worker/registration-events/index.html
jeffposnick: Done.

service-worker/registration-events/index.html
jeffposnick: See previous: https://github.com/GoogleChrome/samples/pull/21#discussion_r19172434

service-worker/registration-events/index.html
jeffposnick: I've changed the first part.

I'm hesitant to include the second bit here, since it's not possible (AFAIK) to trigger a SW update the way the example is set up, so it's not something that the user can try out for themselves.

I think it will make sense to put together a separate sample in which it's easy to trigger SW updates and covering update-specific material there—like this trick, and like an example of clearing out old data in the `onactivate` handler.

service-worker/registration-events/service-worker.js
jeffposnick, line 12: Added something about that.

service-worker/registration-events/service-worker.js
jeffposnick, line 19: Done.

jeffposnick:
@KenjiBaheux @jakearchibald Here are the overdue updates.

The code seems to be working as of Chrome 40.0.2196.0.

jeffposnick:
Also looping in @slightlyoff in case he has any feedback before this gets merged.

service-worker/registration-events/index.html
```diff
@@ -0,0 +1,169 @@
...
+    var liElement = document.createElement('li');
+    liElement.textContent = state;
+    document.querySelector('#states').appendChild(liElement);
+  }
+
+  if ('serviceWorker' in navigator) {
+    // The current browser supports Service Workers.
+    document.querySelector('#availability').innerText = 'are';
+
+    // navigator.serviceWorker.controlled iff there is currently a Service Worker handling
```
dominiccooney: How does this stuff localize? Would it be better to have a table with label: value where the value changes?

service-worker/registration-events/index.html
dominiccooney: Kenji's understanding is right. The registration is really the association between a scope and a number of Service Workers that move through it over time. You can have installing, waiting and active all refer to different ServiceWorker objects simultaneously.

jakearchibald:
LGTM

jeffposnick:
Thanks for the reviews, all.

PR #22, opened 2014-09-16
Consistent, WSK-based styles for all the examples.

jeffposnick:
@addyosmani @PaulKinlan 

Addresses the feedback in #4 across the current set of examples.


jeffposnick:
I'm assuming this is non-controversial (since it follows the pattern from https://github.com/GoogleChrome/samples/pull/14) so I'm going to merge.

PR #25, opened 2014-10-02
First cut of generator samples.

jeffposnick:
@ebidel @addyosmani & co.:

Here's my initial code for a comprehensive example of iterating using generators.

I need to add some explanatory text and there's going to be a complementary H5R article, but I was hoping to get the code reviewed first.

Many of the existing samples related to generators gloss over the gnarly bits, so I deliberately contrived a more complicated use case to show off as many aspects as I could.

Live demo at https://rawgit.com/jeffposnick/samples/es6-generators/generators/index.html

### DIFF
--- a/generators/index.html
+++ b/generators/index.html
@@ -76,7 +76,8 @@ limitations under the License.
 
         // Assuming maxWords isn't set via the generator constructor, this will repeat infinitely.
         // If maxWords is set, then terminate the loop once count is equal to it.
-        while (maxWords === undefined || count < maxWords) {
+        maxWords = maxWords || Infinity;
+        while (count < maxWords) {
           var nextWord = loremIpsumArray[i];
           // uppercase takes its value from the generator constructor the first time through the loop,
           // and takes its value from an optional parameter to .next() each subsequent time.
@@ -100,14 +101,15 @@ limitations under the License.
         }
 
         // If we reach this point due to loop termination, then there's an implicit
-        //   return { done: true, value: undefined };
-        // Be sure to check the .done property if you're iterating with .next()!
-        // If you iterate with a for..of loop, it will automatically terminate when .done is true.
+        //   return undefined;
+        // which in turn yields {done: true, value: undefined}.
+        // If you iterate with a for..of loop, it will automatically terminate when .done is true,
+        // but be sure to check the .done property if you're iterating with .next()!
       }
 
       document.querySelector('#add-many-words').addEventListener('click', function() {
         // This will iterate over an infinite sequence of words.
-        var infiniteIpsumIterator = loremIpsumGenerator();
+        var infiniteIpsumIterator = new loremIpsumGenerator();
         var manyWordsElement = document.querySelector('#many-words');
 
         for (var word of infiniteIpsumIterator) {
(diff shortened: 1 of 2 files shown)


generators/index.html
```diff
@@ -0,0 +1,178 @@
...
+          }
+
+          // uppercase is assigned the value passed into the .next() invoked following
+          // the yield, when execution flow switches back to this generator.
+          // It is not assigned a value until another .next() is called and generator execution resumes!
+          // The value is used the subsequent time through the loop.
+          uppercase = yield nextWord;
+        }
+
+        // If we reach this point due to loop termination, then there's an implicit
```
arv, line 103: This is not quite right. There is an implicit`return undefined` which yields a `{done: true, value: undefined}` and ends the generator.

generators/index.html
```diff
@@ -0,0 +1,178 @@
...
+          // Create an iterator from the generator if it doesn't already exist.
+          // The iterator will yield 5 words, and the first word will not be in uppercase.
+          fiveWordIpsumIterator = loremIpsumGenerator(5, false);
+        }
+
+        addNextWord(false);
+      });
+
+      document.querySelector('#add-uppercase-word').addEventListener('click', function() {
+        if (fiveWordIpsumIterator === undefined) {
```
arv: You could move this into addNextWord to reduce the code duplication.

generators/index.html
```diff
@@ -0,0 +1,178 @@
...
+      });
+
+      var fiveWordIpsumIterator;
+      var fiveWordsElement = document.querySelector('#five-words');
+
+      document.querySelector('#add-word').addEventListener('click', function() {
+        if (fiveWordIpsumIterator === undefined) {
+          // Create an iterator from the generator if it doesn't already exist.
+          // The iterator will yield 5 words, and the first word will not be in uppercase.
+          fiveWordIpsumIterator = loremIpsumGenerator(5, false);
```
arv: I actually feel like we should try to evangelize people to use `new LoremIpsumGenerator(5, false)` since it is more correct than using [[Call]]. [[Call]] just does a [[Construct]] for generators.

generators/index.html
```diff
@@ -0,0 +1,178 @@
...
+
+      //The '*' character after 'function' indicates that this is a generator, not a normal function.
+      function* loremIpsumGenerator(maxWords, uppercase) {
+        // Initialize state. Each iterator based on this generator maintains its own state.
+        var count = 0;
+        var i = 0;
+
+        // Assuming maxWords isn't set via the generator constructor, this will repeat infinitely.
+        // If maxWords is set, then terminate the loop once count is equal to it.
+        while (maxWords === undefined || count < maxWords) {
```
arv: or do:

``` js
maxWords = maxWords || Infinity
```

then you can skip the undefined check below.

generators/index.html
jeffposnick, line 103: I've updated the comment.

generators/index.html
jeffposnick: Good point.

generators/index.html
jeffposnick: I've prepended `new` here and elsewhere.

generators/index.html
jeffposnick: Switched.

jeffposnick:
Thanks for the review, @arv!

PR #26, opened 2014-10-03
Fixed a couple of typos, and added explanatory text.

jeffposnick:
Should be fairly uncontroversial, but CC:ing @arv in case he has any thoughts on the explanatory text.

Live version at https://rawgit.com/jeffposnick/samples/more-es6-generators/generators/index.html


arv:
LGTM

PR #27, opened 2014-10-03
Web application manifest sample.

jeffposnick:
@addyosmani @ebidel @PaulKinlan @cwilso (and anyone else who might be interested):

Here's a basic example of a page that uses a `manifest.json` to specify web app capabilities/metadata.

The `manifest.json` is pretty bare right now, so if there's something specific you think it should demonstrate, your feedback is welcome.

You can try it out at https://rawgit.com/jeffposnick/samples/web-application-manifest/web-application-manifest/index.html using Chrome 39+ on Android.


addyosmani:
I think it's worth us fleshing out the `sizes` section to demonstrate multiple icon sizes. Also worth pulling in @mounirlamouri for a review.

jeffposnick:
Any idea which sizes will actually trigger different behavior on Android Chrome right now?

I added entries for everything currently under https://github.com/GoogleChrome/samples/tree/gh-pages/images though I'm not sure which will have any practical effect.

PaulKinlan:
I am not actually sure if we even use icons yet.

jeffposnick:
Chrome 39 on Android will use at least the 192px icon included in the manifest--try with the URL linked to above. Just not sure what scenarios will trigger which alternative icons to be used.

addyosmani:
The `sizes` property was implemented in https://chromium.googlesource.com/chromium/src/+/b847169dd6954fee83c5b89f6ef4699d6ff018f1 for anyone interested in digging further. I think it would be useful to check with Mounir where this change would actually manifest something the user would see differently in the current setup on Android.

web-application-manifest/README.md
```diff
@@ -0,0 +1,5 @@
+Web Application Manifest Sample
+===
+See https://googlechrome.github.io/samples/web-application-manifest/index.html for a live demo.
+
+Learn more at http://www.chromestatus.com/feature/6488656873259008
```
mounirlamouri, line 5: Shouldn't you point to the spec? https://w3c.github.io/manifest/

web-application-manifest/index.html
```diff
@@ -0,0 +1,69 @@
...
+
+    <meta name="description" content="Sample illustrating the use of the Web Application Manifest.">
+
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>Web Application Manifest Sample</title>
+
+    <link rel="manifest" href="manifest.json">
+
+    <link rel="icon" href="../images/favicon.ico">
```
mounirlamouri, line 30: FWIW, this is going to be ignored with regards to "Add to homescreen".

web-application-manifest/index.html
```diff
@@ -0,0 +1,69 @@
...
+    </p>
+
+
+    <p>
+      Try me out in Chrome 39+ on Android: from Chrome's menu, select
+      <strong>Add to homescreen</strong>.
+    </p>
+
+    <p>
+      When you the new icon on your homescreen, I'll launch as a fullscreen app,
```
mounirlamouri: s/fullscreen app/web app/ or standalone app?

web-application-manifest/manifest.json
```diff
@@ -0,0 +1,28 @@
...
+      "type": "image/png"
+    },
+    {
+      "src": "../images/touch/chrome-touch-icon-192x192.png",
+      "sizes": "192x192",
+      "type": "image/png"
+    }
+  ],
+  "start_url": "index.html",
+  "display": "fullscreen",
```
mounirlamouri: I would use 'standalone' because Chrome Android doesn't support 'fullscreen' and fallbacks to 'standalone'. Fullscreen is meant for desktop mostly.

web-application-manifest/manifest.json
```diff
@@ -0,0 +1,28 @@
...
+      "sizes": "128x128",
+      "type": "image/png"
+    },
+    {
+      "src": "../images/touch/ms-touch-icon-144x144-precomposed.png",
+      "sizes": "144x144",
+      "type": "image/png"
+    },
+    {
+      "src": "../images/apple-touch-icon-precomposed.png",
```
mounirlamouri: It is very unfortunate to have "apple-touch-*" icons in that list...

web-application-manifest/manifest.json
```diff
@@ -0,0 +1,28 @@
...
+      "sizes": "152x152",
+      "type": "image/png"
+    },
+    {
+      "src": "../images/touch/chrome-touch-icon-192x192.png",
+      "sizes": "192x192",
+      "type": "image/png"
```
mounirlamouri: It would be great if you could specify the expected densities of the icons. The algorithm being used to pick the right icon use in priorities icons of the right density.

The optimal size is 48x48 for density=1, then you have to multiply by the dpi.

web-application-manifest/README.md
jeffposnick, line 5: We're following the pattern of linking to the chromestatus.com page from each sample, which in turn has links to the spec, as well as implementation status in other browsers, links to other demos, etc.

web-application-manifest/index.html
jeffposnick, line 30: Yep, but it's still needed in order to show the in-browser favicon/shortcut icon, right? There's no default `/favicon.ico` for this domain.

web-application-manifest/index.html
jeffposnick: Switched to standalone app.

web-application-manifest/manifest.json
jeffposnick: Switched.

web-application-manifest/manifest.json
jeffposnick: I'll remove.

web-application-manifest/manifest.json
jeffposnick: So it sounds like I should just have one entry for icons, and link to the 192x192 version with `density: 4`?

Is there any benefit from me manually resizing the 192x192 down to 48x48 and then specifying that resized version with `density: 1`?

Those 144x144 and 152x152 versions were just pre-existing icons, and if there's no benefit from linking to them, I'll just leave them out.

jeffposnick:
Thanks for the feedback,@mounirlamouri! I'm still not sure if the icons are optimal, so any additional feedback with specific examples is welcome.

web-application-manifest/index.html
mounirlamouri, line 30: If there is no /favicon.ico, that's fair enough.

mounirlamouri:
You don't **have** to resize icons and have one icon per density but I think for a sample it would be a good thing to show the best practices. The algorithm will try to find the best icon if it can't find the ideal one but given that Android is expecting a very specific icon size in device pixel, it seems that it would be better if the sample was following that rule for all the popular device densities. I'm not sure if 4 is a common device densities but surely 2 to 3 are, right?

jeffposnick:
Great, thanks for the additional info, @mounirlamouri. [This guide](http://developer.android.com/design/style/iconography.html) clued me in to some of the Android specifics when it comes to launcher icons, and I've updated the sample to use a wider range of densities all based on the original hi-res image. PTAL.

mounirlamouri:
I don't know if 2.5 is a common density - in which case we might want to add it. lgtm either way.

jeffposnick:
I didn't see 2.5x listed in the "Alternative Drawables" section of http://developer.android.com/guide/practices/screens_support.html

I'll merge this PR. Thanks!

Issue #29, opened 2014-10-22
SW Sample: Prefetching Resources During Registration


jeffposnick:
It should demonstrate standard registration and prefetching/caching a list of URLs within the `oninstall` handler.

jeffposnick:
Closed by #39

PR #39, opened 2014-10-29
Initial cut of the prefetch sample.

jeffposnick:
@kenjibaheux @jakearchibald @slightlyoff

Here's my initial attempt at the prefetch SW sample, closing #29. It could use some additional comments/explanatory text, but I wanted to get some feedback on the actual code.

It uses the native Cache API, currently only available in the latest Chrome Canary build. (Woo! No polyfills!) Since `cache.addAll()` isn't yet implemented natively, it performs multiple `fetch()` + `cache.put()` operations.

You can theoretically see a live demo of this code at https://rawgit.com/jeffposnick/samples/sw-prefetch/service-worker/prefetch/index.html but I'm in the process of figuring out why it doesn't work from there (`TypeError: Failed to fetch` for all `fetch()` requests), while it does work when accessed via http://localhost:8000.

### DIFF
--- a/service-worker/prefetch/README.md
+++ b/service-worker/prefetch/README.md
@@ -1,0 +1,5 @@
+Service Worker Sample: Prefetching Resources During Registration
+===
+See https://googlechrome.github.io/samples/service-worker/prefetch/index.html for a live demo.
+
+Learn more at http://www.chromestatus.com/feature/6561526227927040
--- a/service-worker/prefetch/index.html
+++ b/service-worker/prefetch/index.html
@@ -19,11 +19,11 @@ limitations under the License.
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
-    <meta name="description" content="Sample of prefetching resources during Service Worker registration.">
+    <meta name="description" content="Sample of pre-fetching resources during Service Worker registration.">
 
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
-    <title>Service Worker Sample: Prefetching Resources During Registration</title>
+    <title>Service Worker Sample: Pre-fetching Resources During Registration</title>
 
     <!-- Add to homescreen for Chrome on Android -->
     <meta name="mobile-web-app-capable" content="yes">
@@ -47,12 +47,12 @@ limitations under the License.
   </head>
 
   <body>
-    <h1>Service Worker Sample: Prefetching Resources During Registration</h1>
+    <h1>Service Worker Sample: Pre-fetching Resources During Registration</h1>
 
     <p>Available in <a href="http://www.chromestatus.com/feature/6561526227927040">Chrome 40+</a></p>
 
     <p>
-      This sample demonstrates basic Service Worker registration, in conjunction with pre-caching of a list
+      This sample demonstrates basic Service Worker registration, in conjunction with pre-fetching of a list
       of specific resource URLs during the installation phase.
     </p>
 
@@ -60,26 +60,29 @@ limitations under the License.
       Visit <code>chrome://serviceworker-internals</code> and click on the "Inspect" button next to
       the registered Service Worker to view logging statements for the various actions the
       <code><a href="service-worker.js">service-worker.js</a></code> script is performing.
-    </p>
-
-    <p>
-      The resources that have been pre-fetched should be available without a network connection—
-      try disabling your network and then clicking on the corresponding links. Contrast that with
-      the resource that have not been pre-fetched which will not be initially available without a
-      network connection. There's also an example of a non-existent resource that will always
-      return a HTTP 404.
+      Since most of the action takes place during the Service Worker's install handler, and that is
+      only executed the very first time you visit this page (unless the Service Worker script changes),
+      it can be helpful to inspect the logging in a
+      <a href="https://support.google.com/chrome/answer/95464">Chrome Incognito window</a>.
     </p>
 
     <!-- // [START code-block] -->
     <div class="output">
       <div id="status"></div>
 
-      <ul id="files" style="display: none">
-        <li><a href="static/file1.txt" target="_blank">file1.txt (pre-fetched)</a></li>
-        <li><a href="static/file2.txt" target="_blank">file2.txt (pre-fetched)</a></li>
-        <li><a href="static/file3.txt" target="_blank">file3.txt (not pre-fetched)</a></li>
-        <li><a href="static/does_not_exist.txt" target="_blank">does_not_exist.txt (returns a 404)</a></li>
-      </ul>
+      <div id="files" style="display: none">
+        <p>
+          Some of the resources below have been pre-fetched, and some haven't. Try disabling your network connection
+          and clicking on all the links. The resources that have been pre-fetched should be served from the Service
+          Worker's cache and display as expected; the resource that hasn't been pre-fetched will be unavailable.
+        </p>
+
+        <ul>
+          <li><a href="static/pre_fetched.txt" target="_blank">pre_fetched.txt</a></li>
+          <li><a href="static/pre_fetched.html" target="_blank">pre_fecthed.html</a></li>
+          <li><a href="static/not_pre_fetched.txt" target="_blank">not_pre_fetched.txt</a></li>
+        </ul>
+      </div>
     </div>
 
     <script>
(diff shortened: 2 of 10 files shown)


jeffposnick:
(It turns out that rawgit.com responds to URLs that don't need special handling, like https://rawgit.com/jeffposnick/samples/sw-prefetch/service-worker/prefetch/static/file1.txt, with a HTTP 301 to to the corresponding raw.githubusercontent.com. raw.githubusercontent.com doesn't support CORS, and `fetch()` isn't happy with non-CORS requests in the current Chrome Canary.)

jeffposnick:
I've added some additional comments and workarounds.

This code should work as expected when served off of https://googlechrome.github.io but it's currently broken when accessed via the https://rawgit.com URL, due to the bug(?) described at http://crbug.com/429193

service-worker/prefetch/service-worker.js
```diff
@@ -0,0 +1,84 @@
...
+  // All of these logging statements should be visible via the "Inspect" interface
+  // for the relevant SW accessed via chrome://serviceworker-internals
+  console.log('Handling install event. Resources to pre-fetch:', urlsToPrefetch);
+
+  event.waitUntil(
+    caches.open('prefetch-sample').then(function(cache) {
+      // This two-stage fetch() + cache.put() combo is required because Chrome doesn't
+      // currently support cache.addAll(). Once support for that is added, it's preferable
+      // to use that.
+      // TODO: Switch code to use cache.addAll() when available.
```
gauntface: cache.addAll is needed until this can be reviewed

jeffposnick: My understanding is that M40 won't include a native cache.addAll(), so this code to simulate its behavior will be what we end up releasing as part of the official samples.

service-worker/prefetch/README.md
```diff
@@ -0,0 +1,5 @@
+Service Worker Sample: Prefetching Resources During Registration
+===
+See https://googlechrome.github.io/samples/service-worker/prefetch/index.html for a live demo.
+
+Learn more at http://www.chromestatus.com/feature/6561526227927040
```
dominiccooney, line 5: Is it better to link to Is Service Worker ready yet? How do developers interpret the chromestatus.com entry?

service-worker/prefetch/index.html
```diff
@@ -0,0 +1,140 @@
...
+
+    <p>Available in <a href="http://www.chromestatus.com/feature/6561526227927040">Chrome 40+</a></p>
+
+    <p>
+      This sample demonstrates basic Service Worker registration, in conjunction with pre-fetching of a list
+      of specific resource URLs during the installation phase.
+    </p>
+
+    <p>
+      Visit <code>chrome://serviceworker-internals</code> and click on the "Inspect" button next to
```
dominiccooney: FWIW the unregister method works, so you could build buttons in your demo to get people back to a clean state. Alternatively, using Incognito is nice and simple.

service-worker/prefetch/index.html
```diff
@@ -0,0 +1,140 @@
...
+      <div id="files" style="display: none">
+        <p>
+          Some of the resources below have been pre-fetched, and some haven't. Try disabling your network connection
+          and clicking on all the links. The resources that have been pre-fetched should be served from the Service
+          Worker's cache and display as expected; the resource that hasn't been pre-fetched will be unavailable.
+        </p>
+
+        <ul>
+          <li><a href="static/pre_fetched.txt" target="_blank">pre_fetched.txt</a></li>
+          <li><a href="static/pre_fetched.html" target="_blank">pre_fecthed.html</a></li>
```
dominiccooney: Spelling in the link text.

service-worker/prefetch/index.html
```diff
@@ -0,0 +1,140 @@
...
+      function showFilesList() {
+        document.querySelector('#files').style.display = 'block';
+      }
+
+      if ('serviceWorker' in navigator) {
+        navigator.serviceWorker.register('./service-worker.js', {scope: './'}).then(
+          function(registration) {
+            document.querySelector('#status').textContent = 'Service Worker registration succeeded.';
+
+            if (registration.installing) {
```
dominiccooney: Would it overcomplicate this to wrap this up in a function? Like:

```
function waitUntilInstalled(registration) {
  return new Promise(function(resolve, reject) {
    if (!registration.installing) {
      resolve();
      return;
    }
    registration.installing.addEventListener('statechange', function(e) {
      if (e.target.state == 'installed') {
        resolve();
      }
      // (could also detect conditions to reject here)
    });   
  });
}
```

then your example becomes more fluent, `waitUntilInstalled(registration).then(showFilesList);` and you can combine it with registration:

```
navigator.serviceWorker.register(...)
  .then(waitUntilInstalled)
  .then(showFilesList);
  .catch(... "Something went wrong" stuff ...);
```

service-worker/prefetch/service-worker.js
```diff
@@ -0,0 +1,84 @@
+self.addEventListener('install', function(event) {
+  var urlsToPrefetch = [
+    './static/pre_fetched.txt',
+    './static/pre_fetched.html',
+    // This is an image that will be used in pre_fetched.html
+    'https://www.chromium.org/_/rsrc/1302286216006/config/customLogo.gif'
+  ];
+
+  // All of these logging statements should be visible via the "Inspect" interface
```
dominiccooney, line 13: Again, I think it is better to direct people to chrome://inspect , Service Workers unless there is functionality that is only available in serviceworker-internals and they must go there.

service-worker/prefetch/README.md
jeffposnick, line 5: We've been following this format with all of the GoogleChrome samples—driving traffic to chromestatus.com so that folks can read the specs as well as get links to the docs and demos.

Right now there's some extra noise on the chromestatus.com page, but we'll clean it up and ensure that there are links to the definitive isserverworkerready page as well as to the various demos prior to the M40 beta.

service-worker/prefetch/index.html
jeffposnick: It would be unregister + a page refresh, right? Maybe I'll just leave in the Incognito recommendation for this sample.

service-worker/prefetch/index.html
jeffposnick: Thanks.

service-worker/prefetch/service-worker.js
jeffposnick, line 13: Changed.

service-worker/prefetch/index.html
jeffposnick: Thanks, that's much nicer. I'll change it. Part of the fun of getting everyone up to speed with SW will be getting everyone up to speed with Promises :-)

service-worker/prefetch/index.html
```diff
@@ -0,0 +1,143 @@
...
+      // is past the "installing" state.
+      function waitUntilInstalled(registration) {
+        return new Promise(function(resolve, reject) {
+          if (registration.installing) {
+            // If the current registration represents the "installing" Service Worker, then wait
+            // until the installation step (during which the resources are pre-fetched) completes
+            // to display the file list.
+            registration.installing.addEventListener('statechange', function(e) {
+              if (e.target.state == 'installed') {
+                resolve();
```
jakearchibald, line 102: Might want to reject on a state change to `redundant` indicating a failed install

service-worker/prefetch/service-worker.js
```diff
@@ -0,0 +1,84 @@
...
+          return fetch(url, {mode: 'no-cors'}).then(function(response) {
+            console.log('Fetching of', url, 'succeeded. Response is', response);
+
+            return response;
+          });
+        })
+      ).then(function(responses) {
+        // Cache each of the responses, using the URL of the request as the key.
+        return responses.map(function(response, i) {
+          return cache.put(urlsToPrefetch[i], response);
```
jakearchibald: Just use `addAll` from https://github.com/coonsta/cache-polyfill/blob/master/lib/caches.js?

service-worker/prefetch/service-worker.js
```diff
@@ -0,0 +1,84 @@
...
+      console.error('Pre-fetching failed:', error);
+    })
+  );
+});
+
+self.addEventListener('fetch', function(event) {
+  console.log('Handling fetch event for', event.request.url);
+
+  event.respondWith(
+    caches.open('prefetch-sample').then(function(cache) {
```
jakearchibald: Could also use `caches.match` from https://github.com/coonsta/cache-polyfill/blob/master/lib/caches.js

service-worker/prefetch/service-worker.js
jeffposnick: Okay, I had intentionally avoided using the polyfill for these demos when it was heavier-weight and conflicted with `this.cache`, but there's no reason not to update things to use it now.

service-worker/prefetch/index.html
jeffposnick, line 102: Done.

Issue #31, opened 2014-10-22
SW Sample: Basic Pass-through Caching


jeffposnick:
It should demonstrate basic registration, and then an `onfetch` handler that automatically caches all responses. Subsequent requests for the same URL are served from the cache.

Until the native Caches API is available, it should use https://github.com/coonsta/cache-polyfill

jeffposnick:
Closed by #40

Issue #32, opened 2014-10-22
SW Sample: Caching of Selective Resources


jeffposnick:
It should demonstrate basic registration, and then an `onfetch` handler that selectively caches responses based on multiple criteria:
- Request URL pattern (e.g. only cache URLs hosted on `example.com`)
- Response MIME type (e.g. only cache `image/*` resposnes). Subsequent requests for the same URL are served from the cache.
- Any other ideas?

Until the native Caches API is available, it should use https://github.com/coonsta/cache-polyfill

gauntface:
- Check the response code is 200.
- Cache Google Fonts.
- Only cache resources requested from pages that we have cached during the install step.

PR #40, opened 2014-11-05
Pass-through caching demo.

jeffposnick:
@kenjibaheux @jakearchibald @slightlyoff & co.:

Here's my initial attempt at a pass-through caching example, closing #31.

You can see a live demo of this code at https://rawgit.com/jeffposnick/samples/sw-pass-through-caching/service-worker/pass-through-caching/index.html

### DIFF
--- a/service-worker/pass-through-caching/index.html
+++ b/service-worker/pass-through-caching/index.html
@@ -1,0 +1,122 @@
+<!doctype html>
+<!--
+Copyright 2014 Google Inc. All Rights Reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+    <meta name="description" content="Sample of pass-through caching with Service Workers.">
+
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>Service Worker Sample: Pass-through Caching</title>
+
+    <!-- Add to homescreen for Chrome on Android -->
+    <meta name="mobile-web-app-capable" content="yes">
+    <link rel="icon" sizes="192x192" href="../../images/touch/chrome-touch-icon-192x192.png">
+
+    <!-- Add to homescreen for Safari on iOS -->
+    <!-- TODO: Replace PLACEHOLDER with feature name. -->
+    <meta name="apple-mobile-web-app-title" content="Service Worker Sample: Pass-through Caching">
+
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-status-bar-style" content="black">
+    <link rel="apple-touch-icon-precomposed" href="../../images/apple-touch-icon-precomposed.png">
+
+    <!-- Tile icon for Win8 (144x144 + tile color) -->
+    <meta name="msapplication-TileImage" content="images/touch/ms-touch-icon-144x144-precomposed.png">
+    <meta name="msapplication-TileColor" content="#3372DF">
+
+    <link rel="icon" href="../../images/favicon.ico">
+
+    <link rel="stylesheet" href="../../styles/main.css">
+  </head>
+
+  <body>
+    <h1>Service Worker Sample: Pass-through Caching</h1>
+
+    <p>Available in <a href="http://www.chromestatus.com/feature/6561526227927040">Chrome 40+</a></p>
+
+    <p>
+      This sample demonstrates basic Service Worker registration, in conjunction with pass-through
+      caching. After the Service Worker starts controlling this page, the first time a specific
+      resource is requested, it's <code>fetch()</code>ed from the network and a copy of the response
+      is stored in the Service Worker's cache. All subsequent times that the same resource is
+      requested, it's returned directly from the cache.
+    </p>
+
+    <p>
+      Visit <code>chrome://serviceworker-internals</code> and click on the "Inspect" button next to
+      the registered Service Worker to view logging statements for the various actions the
+      <code><a href="service-worker.js">service-worker.js</a></code> script is performing.
+    </p>
+
+    <!-- // [START code-block] -->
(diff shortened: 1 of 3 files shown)


service-worker/pass-through-caching/service-worker.js
```diff
@@ -0,0 +1,37 @@
...
+          console.log(' Found response in cache:', response);
+
+          return response;
+        } else {
+          // Otherwise, if there is no entry in the cache for event.request, response will be
+          // undefined, and we need to fetch() the resource.
+          console.log(' No response for %s found in cache. About to fetch from network...', event.request.url);
+
+          return fetch(event.request).then(function(response) {
+            console.log('  Response for %s from network is: %O', event.request.url, response);
```
gauntface, line 27: I'd recommended checking the response was valid AND that you got a response.status == 200. If either of those are broken, then bail.

jeffposnick, line 27: What would checking that the response is valid entail—just a null check (are there any cases in which `fetch()` resolves with null instead of rejecting?) or is there some specific property of response that should be checking? I see that there's a [termination reason](https://fetch.spec.whatwg.org/#concept-response-termination-reason) defined as part of the response object, but I don't actually see it mapped to a property of what gets retruned in the current Chrome implementation.

Additionally, since Chrome returns [opaque filtered responses](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque) for non-CORS requests, it's possible that `response.status == 0` is a legitimate, cachable response. So it's a tradeoff between never caching opaque responses or caching them unconditionally, and risking caching a transient error response.

That being said, I will at least check `response.status >= 400`, to avoid caching any non-opaque error responses.

service-worker/pass-through-caching/index.html
```diff
@@ -0,0 +1,122 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>Service Worker Sample: Pass-through Caching</title>
+
+    <!-- Add to homescreen for Chrome on Android -->
+    <meta name="mobile-web-app-capable" content="yes">
+    <link rel="icon" sizes="192x192" href="../../images/touch/chrome-touch-icon-192x192.png">
+
+    <!-- Add to homescreen for Safari on iOS -->
+    <!-- TODO: Replace PLACEHOLDER with feature name. -->
```
dominiccooney, line 30: PLACEHOLDER seems to have been removed?

service-worker/pass-through-caching/service-worker.js
```diff
@@ -0,0 +1,47 @@
...
+            if (response && response.status < 400) {
+              // This avoids caching responses that we know are errors (i.e. HTTP status code of 4xx or 5xx).
+              // One limitation is that, for non-CORS requests, we get back a filtered opaque response
+              // (https://fetch.spec.whatwg.org/#concept-filtered-response-opaque) which will always have a
+              // .status of 0, regardless of whether the underlying HTTP call was successful. Since we're
+              // blindly caching those opaque responses, we run the risk of caching a transient error response.
+              //
+              // We need to call .clone() on the response object to save a copy of it to the cache.
+              // (https://fetch.spec.whatwg.org/#dom-request-clone)
+              cache.put(event.request, response.clone());
```
dominiccooney, line 38: Over time the cache could get quite plump. Is it worth noting that storage isn't unlimited and at least mention that you should have some strategy to get rid of old stuff?

service-worker/pass-through-caching/service-worker.js
gauntface, line 27: I was discussing this point with a colleague and we both ended up agreeing that caching only 200 and 0 responses would be the safest bet for a recommended sample.

It will save people a lot of pain should a bad response get cached.

service-worker/pass-through-caching/index.html
jeffposnick, line 30: Yes, thanks for catching that.

service-worker/pass-through-caching/service-worker.js
jeffposnick, line 38: Thanks, that's very true.

What's the limitation on Cache API storage that we're planning on enforcing with Chrome 40? And is that limit enforced based on cache name, service worker registration, or something else?

I'd like to include some actual numbers based on the Chrome implementation, and will add the caveat that other browsers or future versions of Chrome might enforce different limits.

I'll also add something about how naive pass-through caching is appropriate for web apps that make a limited set of HTTP requests, but web apps that might end up making arbitrary HTTP requests (like, say, an RSS client or news client) shouldn't blindly pass-through cache everything and need to periodically prune their caches if they do cache aggressively.

service-worker/pass-through-caching/service-worker.js
jeffposnick, line 27: I talked to @slightlyoff about this, and he was (somewhat?) surprised that opaque filtered responses always had a `status` of 0, and suggested that I file https://github.com/whatwg/fetch/issues/14 to see whether that restriction could be relaxed in the `fetch()` spec.

I'll keep caching whenever `response.status < 400` and make a point of specifically highlighting this tradeoff in the comments. As long as the `fetch()` spec stays the same, it'll be each developer's decision whether they want to err on the side of caching too much or too little.

service-worker/pass-through-caching/service-worker.js
```diff
@@ -0,0 +1,47 @@
...
+          // and we can just return it.
+          console.log(' Found response in cache:', response);
+
+          return response;
+        } else {
+          // Otherwise, if there is no entry in the cache for event.request, response will be
+          // undefined, and we need to fetch() the resource.
+          console.log(' No response for %s found in cache. About to fetch from network...', event.request.url);
+
+          return fetch(event.request).then(function(response) {
```
jakearchibald: Clone request too, as `fetch` consumes it (as does `cache.put`)

service-worker/pass-through-caching/service-worker.js
jakearchibald, line 38: > naive pass-through caching is appropriate for web apps that make a limited set of HTTP requests

Although apps that make a limited set of requests should probably be caching them up front in the install step

jeffposnick, line 38: I've filed https://github.com/GoogleChrome/samples/issues/42 to track the fact that this sample and probably a few others will need to address cache maintenance/versioning. I'll make those changes across all the samples in a consistent fashion in a follow-up PR.

Agreed that in general, pre-fetching will be a best practice, and https://github.com/GoogleChrome/samples/issues/32 will cover more selective caching scenarios. I can envision there being some use cases where blindly caching everything is appropriate, though I'll make it clear in the comments that developers need to clearly understand their use case.

jeffposnick:
So I've added some caveats to explain that this particular aggressive form of caching is not appropriate for many use cases. I am going to merge this now since I think it's still useful as a sample, but the other samples we put out might be more relevant for real-world usage.

PR #57, opened 2014-12-05
Fix unregister function

cesarwbr:
Hi!
Following the specification you can only unregister a Service Worker register through a ServiceWorkerRegistration.
I just tested it on Chrome 40 beta and it definitely works.

### DIFF
--- a/service-worker/basic/index.html
+++ b/service-worker/basic/index.html
@@ -16,12 +16,14 @@ limitations under the License.
 -->
 <script>
 var scope = 'hello/';
+var registration;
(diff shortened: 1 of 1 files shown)


jeffposnick:
Deferring to @paulirish as it's his sample.

paulirish:
Can you switch lines 25 & 26? Makes it a touch more readable.

Otherwise this looks great. Thanks for the PR!

cesarwbr:
Yes @paulirish.

Done!

PR #58, opened 2014-12-15
SW Sample: Offline Analytics

jeffposnick:
@jakearchibald @ebidel & co.:

This builds on an existing sample that demonstrates offline functionality, and addresses the question of how to get a system like Google Analytics that relies on sending data to a remote server to play nicely with offline web pages.

The code queues failed Google Analytics pings in `IndexedDB` and attempts to resend them each time the service worker starts up, with the `dt=` parameter set to reflect the original time of the page view event.

### DIFF
--- a/service-worker/offline-analytics/service-worker.js
+++ b/service-worker/offline-analytics/service-worker.js
@@ -1,0 +1,188 @@
+/*
+ Copyright 2014 Google Inc. All Rights Reserved.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+// While overkill for this specific sample in which there is only one cache,
+// this is one best practice that can be followed in general to keep track of
+// multiple caches used by a given service worker, and keep them all versioned.
+// It maps a shorthand identifier for a cache to a specific, versioned cache name.
+
+// Note that since global state is discarded in between service worker restarts, these
+// variables will be reinitialized each time the service worker handles an event, and you
+// should not attempt to change their values inside an event handler. (Treat them as constants.)
+
+// If at any point you want to force pages that use this service worker to start using a fresh
+// cache, then increment the CACHE_VERSION value. It will kick off the service worker update
+// flow and the old cache(s) will be purged as part of the activate event handler when the
+// updated service worker is activated.
+var CACHE_VERSION = 1;
+var CURRENT_CACHES = {
+  'offline-analytics': 'offline-analytics-v' + CACHE_VERSION
+};
+
+var idbDatabase;
+var STORE_NAME = 'urls';
+
+function openDatabaseAndReplayRequests() {
+  var indexedDBOpenRequest = indexedDB.open('offline-analytics', CACHE_VERSION);
+
+  indexedDBOpenRequest.onerror = function(error) {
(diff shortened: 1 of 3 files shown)


service-worker/offline-analytics/service-worker.js
```diff
@@ -0,0 +1,221 @@
...
+self.addEventListener('activate', function(event) {
+  // Delete all caches that aren't named in CURRENT_CACHES.
+  // While there is only one cache in this example, the same logic will handle the case where
+  // there are multiple versioned caches.
+  var expectedCacheNames = Object.keys(CURRENT_CACHES).map(function(key) {
+    return CURRENT_CACHES[key];
+  });
+
+  event.waitUntil(
+    caches.keys().then(function(cacheNames) {
```
ebidel, line 130: I'm a noob, `caches` is global?

service-worker/offline-analytics/service-worker.js
```diff
@@ -0,0 +1,221 @@
...
+              // One limitation is that, for non-CORS requests, we get back a filtered opaque response
+              // (https://fetch.spec.whatwg.org/#concept-filtered-response-opaque) which will always have a
+              // .status of 0, regardless of whether the underlying HTTP call was successful. Since we're
+              // blindly caching those opaque responses, we run the risk of caching a transient error response.
+              //
+              // We need to call .clone() on the response object to save a copy of it to the cache.
+              // (https://fetch.spec.whatwg.org/#dom-request-clone)
+              cache.put(event.request, response.clone());
+            } else {
+              // If this is a Google Analytics ping then we want to retry it if a HTTP 4xx/5xx response
```
ebidel: Is a 500 something one should retry? If that URL continues to 500, won't this continue to be retried every time the SW starts?

ebidel:
This is an awesome use case! I'm super excited about this. It's like we're actually making offline happen :)

cc @slightlyoff

service-worker/offline-analytics/service-worker.js
ciromaciel, line 130: http://www.youtube.com/watch?v=4uQMl7mFB6g&t=5m0s

ebidel, line 130: This was more of a suggestion to add a comment :)

service-worker/offline-analytics/service-worker.js
jeffposnick: I think retrying on a Google Analytics 5xx response is the right thing to do, as it's hard to imagine a GA returning a 5xx for a permanent error.

After thinking about it some more, I'm less inclined to retry on 4xx responses. I've seen situations in which Google's web servers return a transient 403 when DOS protection logic kicks in, but that's pretty much the only scenario I could imagine in which retrying later on would work, and detecting that specifically would be tough.

Sound reasonable if I switch the logic to retry on all 5xx, and don't retry for 4xx?

ebidel: GG. To generalize, what if we had a `NUM_RETRIES` that users could configure? Set the default to 1 or 2. You'd have to save that alongside the idb entry yea? Thoughts?

jeffposnick: Agreed that there should be some sort of "I give up" logic in the retry.

Since this is a Google Analytics-specific example, and since GA pings that are re-sent after too much time has passed are effectively worthless, I'm more inclined to put in logic that gives up if more than, say, 1 day has passed. (The [docs](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#qt) say anything older than 4 hours might be ignored, but in practice it might be longer.)

(The actual number of times that the request might be retried over the course of the day will vary depending on whether there's anything automated that revives the service worker script in the background or how frequently the user visits a controlled page.)

service-worker/offline-analytics/service-worker.js
jeffposnick, line 130: Added a comment.

jeffposnick:
Thanks for the review. PTAL.

ebidel:
:+1:

PR #64, opened 2015-01-16
adds 'any-pointer' and 'any-hover' media queries sample

samthor:


media-hover-pointer/index.html
```diff
@@ -0,0 +1,214 @@
...
+      </style>
+      <div class="output">
+        <button id="sized_1" class="sized">
+          Click Me
+        </button>
+        <label for="sized_1">This button will be larger where there is no fine pointer available</label>
+      </div>
+    <!-- // [END code-block] -->
+
+    <!-- // [START code-block] -->
```
jeffposnick: Can you use a different name for this block? If we decide to auto-include this content into an updates.html5rocks.com article we'd need to distinguish between the two named blocks of code.

jeffposnick:
LGTM with the one small comment. Thanks!

media-hover-pointer/index.html
samthor: Done!

RByers:
It would be nice if the sample reported exactly which queries were true on the device in question (so I can see exactly what it means for me instead of just a couple examples).  Eg. we've been using this test page: http://jsbin.com/heniko/.

media-hover-pointer/index.html
```diff
@@ -0,0 +1,214 @@
...
+    <p>
+      
+    </p>
+
+    <!-- // [START code-block-any-pointer] -->
+    <style>
+      .sized {
+        font-size: 0.5em;
+      }
+      @media (any-pointer: none), (any-pointer: coarse) {
```
RByers, line 69: The text says "This button will be larger where there is no fine pointer available" but that's not what this MQ does (eg. try it on a chromebook pixel, the button is big even though there is a mouse - because 'any-pointer: fine' and 'any-pointer: coarse' are both true).  Maybe you want "not (any-pointer: fine)" instead?

PR #65, opened 2015-01-16
webaudio demo

samthor:
This isn't too complex, although I'm kind of unsure whether playing a boring single tone is interesting enough. I'd have to commit some audio to the repo otherwise.


jeffposnick:
Can you use the Page Visibility API ([demo](http://simpl.info/pagevisibility/)) to detect whether the tab is focused?

samthor:
I've updated the demo. PTAL.

jeffposnick:
:+1:

PR #66, opened 2015-01-19
web-animations demos (for m36+)

samthor:
This is a demo for the older Web Animations feature (at least `element.animate()`, added in m36).


jeffposnick:
:+1: Thanks!

PR #68, opened 2015-01-26
Add samples for ES6 Computed Property Names

addyosmani:
Adds initial samples for [computed property names](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-object-initializer-runtime-semantics-evaluation), which work behind a flag in 42.0.2288.0 and above.

Note: I was unable to find a Chromium dashboard entry for computed properties. I assume this is a minor oversight or we plan to fold them under a broader "Object Literal Enhancements" entry. Commented on https://code.google.com/p/v8/issues/detail?id=3754#c11.

cc @arv in case he has any feedback.

### DIFF
--- a/computed-properties-es6/index.html
+++ b/computed-properties-es6/index.html
@@ -55,7 +55,7 @@ limitations under the License.
         similar to <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors">property accessors</a>.
         Below, computed property names are demonstrated using Object Literals, ES6 Classes and ES6 Symbols.
     </p>
-    <p>This feature is vailable in Chrome 42 behind the <code>--harmony-computed-property-names</code> flag. For step-by-step instructions on how to use command-line
+    <p>This feature is available in Chrome 42 behind the <code>--harmony-computed-property-names</code> flag. For step-by-step instructions on how to use command-line
     flags with Chrome, read the <a href="http://www.chromium.org/for-testers/command-line-flags">command-line flags</a> guide. In this case, you would
     launch Chrome with <code>--args --js-flags="--harmony-computed-property-names"</code>.</p>
 
--- a/computed-properties-es6/index.html
+++ b/computed-properties-es6/index.html
@@ -99,32 +99,25 @@ limitations under the License.
         // -> A method with a computed name was called
 
         // Example 3
-        log('\nUsing Computed property names with ES6 Classes:');
-        function ID(x) {
-            return x;
-        }
-
-        class Person {
-            getUsername() { return 'ErikArvidsson'}
-            ['getLastName']() { return 'Arvidsson'; }
-            getGroup() { return 'TC39'; }
-            [ID('getTeam')]() { return 'Chrome'; }
-        }
-        log(new Person().getUsername());
-        log(new Person().getLastName());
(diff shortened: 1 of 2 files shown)


computed-properties-es6/index.html
```diff
@@ -0,0 +1,139 @@
...
+        var O = {
+                [fieldName]: "value",
+                [methodName]() { log("A method with a computed name was called"); }
+        };
+        O[methodName]();
+        // -> A method with a computed name was called
+
+        // Example 3
+        log('\nUsing Computed property names with ES6 Classes:');
+        function ID(x) {
```
arv: This is not very useful/illustrative. The V8 unit test has this to ensure that non literals work in that context.

arv:
The initial plan was to include computed properties with harmony object literals. Due to time constraints computed properties were not ready so we created a new V8 feature flag. Now that we unshipped classes we might want to enable computed properties when we enable classes.

computed-properties-es6/index.html
```diff
@@ -0,0 +1,139 @@
...
+    <h1>Computed Property Names (ES6) Sample</h1>
+
+    <!--
+    <p>Available in <a href="http://www.chromestatus.com/feature/PLACEHOLDER">Chrome PLACEHOLDER+</a></p>
+    -->
+    <p>ES6 Computed property names allow you to dynamically create property names using an expression in brackets (<code>[ ]</code>)
+        similar to <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors">property accessors</a>.
+        Below, computed property names are demonstrated using Object Literals, ES6 Classes and ES6 Symbols.
+    </p>
+    <p>Available in Chrome 42 behind the <code>--harmony-computed-property-names</code> flag</p>
```
jeffposnick: I'm familiar with flipping chrome://flags/#enable-javascript-harmony, but that's not sufficient to use computed property names.

Can you link to http://www.chromium.org/for-testers/command-line-flags to provide step-by-step instructions to folks who are similarly unfamiliar?

addyosmani:
Thank you for sharing the status update @arv. I think computed props would compliment our class story in M42 well. Are there any remaining blockers on the implementation? Although it's low-pri, happy to help with the outstanding chromestatus entry if we still need one.

addyosmani:
Agreed that the Classes example was a poor illustration. I've folded it into the Symbols example as I think it makes more sense to show computed method names in that context.

computed-properties-es6/index.html
```diff
@@ -0,0 +1,134 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
```
arthurvr, line 20: here you use 2 spaces indentation, inside `<body>` I see 4, and inside the `<script>` some stuff uses 8.

addyosmani:
Landing. I'm happy to refresh/update further when we know if V8 are going to ship computed property names at the same time as classes.

RatStar811 (approved):

RatStar811 (approved):

PR #72, opened 2015-02-10
Replace usage of innerText with textContent

### LINKED ISSUE #71: innerText being used instead of textContent in a couple of places
Specifically, here: https://github.com/GoogleChrome/samples/search?utf8=%E2%9C%93&q=innerText

A+ that it works in Chrome regardless, but for those of us creepin' via FF...we'd sure <3 textContent instead.

brittanydionigi:
Resolves #71, unless there is some specific reason for using innerText in these scenarios that I'm unaware of...

### DIFF
--- a/battery-status/index.html
+++ b/battery-status/index.html
@@ -70,16 +70,16 @@ limitations under the License.
   var warningEl = document.getElementById("warning");
 
   var updateBatteryUI = function(battery) {
-    levelEl.innerText = (battery.level * 100) + "%";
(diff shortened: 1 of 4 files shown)


addyosmani:
Thanks Brittany!

For context: Whilst the performance characteristics of innerText and textContent result in ~ the same speeds in Chrome, compat wise it makes more sense to switch to textContent here. iT was introduced in the IE era.

PR #73, opened 2015-02-12
Custom offline page

jeffposnick:
@slightlyoff: This is an example of using a service worker to respond to failed GET requests for HTML documents with a custom "offline" error page.

There's a live demo at https://rawgit.com/jeffposnick/samples/offline-detection/service-worker/custom-offline-page/index.html

(I wasn't able to get any of the `<meta>` tags to control caching to work consistently, so I resorted to cache-busting.)


slightlyoff:
Awesome!

Can we add some minimal styling and a cute brand icon so that folks taking this off-the-shelf don't need to think hard?

slightlyoff:
Also, can you add a manifest to the page so it can serve as an installability demo? Thanks!

slightlyoff:
LGTM++

PR #75, opened 2015-02-13
adding support information for Firefox to registration-events sample

chrisdavidmills:


jeffposnick:
Hello @chrisdavidmills—I'm really excited about Firefox support for service workers!

This sample is reporting that `navigator.serviceWorker` isn't defined when I tried it under the standard OS X Firefox 35 build. Is service worker support currently limited to some combination of flags/special builds of Firefox? If so, it's probably too early to flag these samples as being supported in Firefox.

Also, a bit of housekeeping—I don't see a Google Contributors License Agreement on record for your GitHub account. We'd need you to [sign it electronically](https://cla.developers.google.com/) before we could merge.

chrisdavidmills:
it's in Fx, but currently behind a pref. Have you got the SW stuff preffed on in Chrome by default now? As of which version? I'll get that updated on MDN.

I'll sign the agreement for future contributions. Cheers Jeff!

jeffposnick:
Yep, Chrome 40+ has core service worker functionality enabled by default, without any flag-flipping required.

jpmedley:
Chris,

I've been marking that on pages as I go.

Joe Medley | Technical Writer | jmedley@google.com | 816-678-7195

jeffposnick:
In general, we've been tagging samples with the earliest version of Chrome that's supported, and linking to the corresponding chromestatus.com entry, which includes more detailed cross-browser compatability info.

PR #80, opened 2015-02-26
Add ES6 Classes samples

addyosmani:
reviewers: @jeffposnick
Also welcome feedback from: @rauschma

Limitations noted in the intent-to-ship, which impact the scope of what was implemented here:

> - At this time we do not support subclassing built-ins and DOM objects (I've added a sample, but commented out)
> - Current ES6 specification (in particular, [[CreateAction]] hook that takes arguments) makes design of subclassable exotic (in particular DOM) objects unclear. We intend to work with involved parties on resolving this difficulty, but for now we add an extra limitation to calls to super constructor in subclasses: if a subclass constructor calls a superclass constructor via ‘super(...)’ call, that call must be the first statement of constructor’s body, and arguments to ‘super(...)’ call can not refer to ‘this’. 
>   We intend to lift this restriction in the future, once we clarify the design for exotic objects’ subclassing.

Note: these samples are intentional minimal to lower the friction involved in getting beginners on-board with the syntax.

### DIFF
--- a/classes-es6/index.html
+++ b/classes-es6/index.html
@@ -1,0 +1,246 @@
+<!doctype html>
+<!--
+Copyright 2014 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+    <meta name="description" content="Sample illustrating the use of Classes (ES6).">
+
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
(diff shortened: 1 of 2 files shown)


classes-es6/index.html
```diff
@@ -0,0 +1,246 @@
...
+
+        let inst = new MyPoly();
+        inst.getPolyName();
+
+
+        // Example 3: Extending an existing class
+        // ===============================================================
+
+        // Classes support extending other classes, but can also extend
+        // other objects. Let's extend the Polygon class to create a new
```
rauschma: Whatever you extend must be a constructor.

classes-es6/index.html
```diff
@@ -0,0 +1,246 @@
...
+        // derived class called Square.
+        class Square extends Polygon {
+            constructor(length) {
+                // The reserved 'super' keyword allows access to parent methods.
+                // Here, it will call the parent class' constructor with lengths
+                // provided for the Polygon's width and height
+                super(length, length);
```
rauschma, line 147: Note: `super()` for making super-constructor calls and `super.prop` for accessing super-properties and making super-method calls are quite different. But maybe that’s too subtle to be explained here.

Possibly: Mention what the default constructor for derived classes looks like.

jeffposnick:
I'll definitely defer to @rauschma on the technical content. Nothing specific to add otherwise on my end. :+1:

addyosmani:
Thanks for the reviews @rauschma and @jeffposnick. I'll get the comments addressed shortly.

Side: It looks like built-ins may in fact work, however the V8 unit tests appear to show limited support. I'll follow up with the team and update the samples here accordingly.

PR #81, opened 2015-02-26
Use 2015 in the placeholder-name template.

jeffposnick:
R: @addyosmani


addyosmani:
LGTM. Let's merge this in after the latest M42 samples so they don't require rebasing.

PR #82, opened 2015-02-27
Add Fetch API demos

addyosmani:
reviewers: @gauntface @jeffposnick

### DIFF
--- a/fetch-api/fetch-post.html
+++ b/fetch-api/fetch-post.html
@@ -1,0 +1,119 @@
+<!doctype html>
+<!--
+Copyright 2014 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+    <meta name="description" content="Sample illustrating the use of Fetch API.">
+
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <title>Fetch API POST Sample</title>
+
+    <!-- Add to homescreen for Chrome on Android -->
+    <meta name="mobile-web-app-capable" content="yes">
+    <link rel="icon" sizes="192x192" href="../images/touch/chrome-touch-icon-192x192.png">
+
+    <!-- Add to homescreen for Safari on iOS -->
+    <meta name="apple-mobile-web-app-title" content="Fetch API Reddit Sample">
+
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-status-bar-style" content="black">
+    <link rel="apple-touch-icon-precomposed" href="../images/apple-touch-icon-precomposed.png">
+
+    <!-- Tile icon for Win8 (144x144 + tile color) -->
+    <meta name="msapplication-TileImage" content="images/touch/ms-touch-icon-144x144-precomposed.png">
+    <meta name="msapplication-TileColor" content="#3372DF">
+
+    <link rel="icon" href="../images/favicon.ico">
+
+    <link rel="stylesheet" href="../styles/main.css">
+</head>
+
+<body>
+<h1>Fetch API POST Sample</h1>
+
+<p>Available in <a href="http://www.chromestatus.com/feature/6730533392351232">Chrome 42+</a></p>
+
+<div class="output">
+    <p id="log"></p>
+</div>
+
+<!-- // [START code-block] -->
+<script>
+    "use strict";
+
+    var logEl =  document.querySelector('#log');
+
+    function log() {
+        logEl.innerHTML += '<p>' + Array.prototype.join.call(arguments, '') + '<p>';
+    }
+
+    function serialize(data) {
+        if(data) {
+            return Object.keys(data).map(function (key) {
+                return encodeURIComponent(key) + '=' + encodeURIComponent(data[key])
+            }).join('&');
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


gauntface:
LGTM. Think samples are solid - are you still struggling with the POST request or is your one working?

Realised my sample is bust thanks for a change in Chrome such that no-cors mode has been disabled due to a security concern.

fetch-api/fetch-post.html
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                log('String: "' + json + '" is not valid JSON');
+                log(data.error);
+            }
+        }).catch(function (error) {
+            log(error.message);
+        });
+    }
+
+    var jsonValidator = 'http://validate.jsontest.com/';
```
jeffposnick: You'll get a mixed content failure when the sample page is accessed via HTTPS.

`Mixed Content: The page at 'https://rawgit.com/GoogleChrome/samples/fetch/fetch-api/fetch-post.html' was loaded over HTTPS, but requested an insecure resource 'http://validate.jsontest.com/'. This request has been blocked; the content must be served over HTTPS.`

And it looks like https://validate.jsontest.com/ isn't a thing.

fetch-api/fetch-reddit.html
```diff
@@ -0,0 +1,118 @@
...
+                        logEl.innerHTML = '<ul>' + links + '</ul>';
+                    });
+        }
+    }
+
+    fetch('https://www.reddit.com/reddits.json')
+            .then(function(response) {
+                return response.json()
+            }).then(function(json) {
+                console.log(json);
```
jeffposnick: Did you intend to keep this, or was it for debugging during development?

fetch-api/fetch-response-metadata.html
```diff
@@ -0,0 +1,88 @@
...
+<!-- // [START code-block] -->
+<script>
+    "use strict";
+
+    function log() {
+        document.querySelector('#log').textContent += Array.prototype.join.call(arguments, '') + '\n';
+    }
+
+    fetch('users.json').then(function(response) {
+        console.log(response);
```
jeffposnick: Did you intend to keep this, or was it for debugging during development?

fetch-api/fetch-reddit.html
```diff
@@ -0,0 +1,118 @@
...
+                        var links = "";
+                        for(var i = 0; i < body.data.children.length; i++){
+                            links += '<li><a href="' + body.data.children[i].data.url + '">' + body.data.children[i].data.url + '</a></li>';
+                        }
+                        logEl.innerHTML = '<ul>' + links + '</ul>';
+                    });
+        }
+    }
+
+    fetch('https://www.reddit.com/reddits.json')
```
jeffposnick: This returned some subreddits of questionable taste when I ran the sample.

I'd recommend an API call that will give a more curated response, like http://www.reddit.com/api/subreddits_by_topic.json?query=javascript

fetch-api/fetch-response-metadata.html
```diff
@@ -0,0 +1,88 @@
...
+        log(response.headers.get('Content-Type'));
+        log(response.headers.get('Date'));
+
+        log(response.status);
+        log(response.statusText);
+        log(response.type);
+        log(response.url);
```
jeffposnick, line 69: Move this to the top of the log.

fetch-api/fetch-response-metadata.html
```diff
@@ -0,0 +1,88 @@
...
+<script>
+    "use strict";
+
+    function log() {
+        document.querySelector('#log').textContent += Array.prototype.join.call(arguments, '') + '\n';
+    }
+
+    fetch('users.json').then(function(response) {
+        console.log(response);
+        log(response.headers.get('Content-Type'));
```
jeffposnick: Could you add the string `'Content-Type: '` prior to logging the value? Ditto for the other lines. It would make it clearer what was being logged without needing to match up lines in the `view-source:`.

fetch-api/fetch-success-error-handlers.html
```diff
@@ -0,0 +1,98 @@
...
+        } else {
+            return Promise.reject(new Error(response.statusText))
+        }
+    }
+
+    function json(response) {
+        return response.json()
+    }
+
+    fetch('users.json')
```
jeffposnick, line 77: How about adding in another, failing `fetch()` call with the same logic? Request `'invalid.json'` or something.

gauntface:
Mega nit pick - any chance of a index.html to make it easier to navigate these demos or will gh-pages do this?

fetch-api/fetch-reddit.html
addyosmani: That makes sense. I'll update this to provide a slightly more predictable quality of results. Thanks Jeff!

fetch-api/fetch-response-metadata.html
addyosmani, line 69: SGTM

fetch-api/fetch-reddit.html
addyosmani: I intended to keep this to demonstrate the complete payload, but didn't want to log to the page. If it's better to drop these to be consistent with other demos happy to do so.

addyosmani:
@gauntface I thought about that, but added in a README index that we'll be pointing to which lists up to all the samples instead. See: https://github.com/GoogleChrome/samples/blob/823dd3edc6816c6a1bd5b1bfba1a69568ec64f28/fetch-api/README.md. We can easily turn that into an index if needed though.

fetch-api/fetch-success-error-handlers.html
addyosmani, line 77: Good idea!

fetch-api/fetch-response-metadata.html
addyosmani: Good call. That will make it more clear what's being output. It'll land with the next set of commits.

gauntface:
Looks good.

fetch-api/fetch-post.html
addyosmani: I understand the mixed content warning, but it looks like the validation service I was using is now down. I'll see if I can find a suitable replacement that we can post data to without mixed content warnings.

addyosmani:
@jeffposnick Addressed your feedback. I've also refactored the POST example to use the GitHub Gist API to create a new gist based on user input. It was one of the few open-APIs that supported unauthenticated POST requests :)

jeffposnick:
:+1:

PR #88, opened 2015-03-04
Push Changes - Feedback from John

gauntface:
After John discussed with Michael - seems like the best approach for handling subscriptionIds is to perform regular syncs - the pushsubscriptionchange method doesn't seem to make sense.

### DIFF
--- a/push-messaging-and-notifications/index.html
+++ b/push-messaging-and-notifications/index.html
@@ -74,7 +74,6 @@ limitations under the License.
     <div class="js-curl-command"></div>
     
     <script src="demo.js"></script>
-    <script src="subscription-controller.js"></script>
     <script src="main.js"></script>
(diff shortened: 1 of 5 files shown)


jeffposnick:
So just to make sure I understand the model we're recommending: whenever you visit a page that wants to use push notifications, you register and get back the `subscription` object, and you unconditionally pass it to `sendSubscriptionToServer()`.

And while implementing `sendSubscriptionToServer()` is outside the scope of this sample, the idea would be that it would submit a combination of the subscription id, the subscription endpoint, and... what else? A user identifier? Or a browser identifier? Or both? And how will the server know to expire old subscription entries?

The changes you made look :+1: to me, so I don't think this should delay your merge. Happy to take the discussion off this PR if needed.

PR #94, opened 2015-03-13
Added Service Worker Sample for posting messages from a page that isn't controlled

TalAter:
This new code sample shows how a page that has just registered a Service Worker, but is still not being controlled by it, can use the `ServiceWorkerRegistration` object to post messages to the new Service Worker.

It also shows how to detect whether to use this method, or good old fashioned (old?) `navigator.serviceWorker.controller`.

All the samples provided so far (and all the tutorials I saw) only began communication using postMessage after the page was refreshed... so finding this little gem in the [Service Worker spec](http://www.w3.org/TR/service-workers/#service-worker-registration-interface) was a lifesaver.

### DIFF
--- a/service-worker/install-time-post-message/README.md
+++ b/service-worker/install-time-post-message/README.md
@@ -1,0 +1,5 @@
+Service Worker Sample: Posting Messages During Installation
+===
+See https://googlechrome.github.io/samples/service-worker/install-time-post-message/index.html for a live demo.
+
+Learn more at http://www.chromestatus.com/feature/6561526227927040
--- a/service-worker/install-time-post-message/index.html
+++ b/service-worker/install-time-post-message/index.html
@@ -1,0 +1,182 @@
+<!doctype html>
+<!--
+Copyright 2015 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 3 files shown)


jpmedley:
This looks good, but I have a few concerns. If the install handler calls event.waitUntil(), the message will be blocked. Maybe your explanation should point this out.

Also, if you're trying to get the jump on the install process, it seems like you might be setting yourself up for a race condition. What's the use case for this trick?

TalAter:
Thanks @jpmedley.

I just tried this out with very long installation process inside an event.waitUntil() and couldn't see how the message could be blocked. I'm using `ServiceWorkerRegistration` which is only returned when the promise from registering the service worker completes.

``` javascript
navigator.serviceWorker.register('sw.js').then(function(registration) {
  // registration contains an installed Service Worker here
})
```

Is there a way that I'm missing which might cause a race condition or the message to be blocked?

As for a use case, there are many. The one I'm looking at right now is a script that saves data into the cache. The data to save is present in the page, and not in the service worker code.

This means that without this method, if I want to send the data to the Service Worker so it will cache it, I need to wait until the second time the user visits this page, and the Service Worker is in control of the page.

Here's some [code](https://github.com/TalAter/UpUp/blob/b40ae6743a168d0a55ff920b501e0c444adc1264/src/upup.js#L84) that uses this trick to allow me to place messages into the cache when the user first visits the page.

PR #99, opened 2015-03-31
Use `String#includes` instead of `indexOf`

mathiasbynens:


jeffposnick:
Looks like `String.prototype.includes()` works in Chrome 41+, so that should be safe now. :+1:

PR #100, opened 2015-03-31
Use `const` over `var`

mathiasbynens:


ebidel:
Are we good with `const` now in ES6? Has v8 optimized it? It actually made for slower code a while back: https://code.google.com/p/chromium/issues/detail?id=80149

PR #101, opened 2015-04-01
SW Sample: Immediate Control

### LINKED ISSUE #30: SW Sample: Registration with Immediate Replacement
It should demonstrate standard registration with a technique for immediately elevating the installing SW to be the controlling SW. (I think this would entail forcing the page to refresh, but it would be good to confirm the best practice first.)

jeffposnick:
R: @jakearchibald @gauntface @addyosmani 

This shows off the `skipWaiting()` and `clients.claim()` combo to induce a service worker to take immediate control of a page.

A live version is available at https://rawgit.com/jeffposnick/samples/sw-immediate-control/service-worker/immediate-control/index.html, and it requires Chrome 42 to make use of both methods.

Fixes #30

### DIFF
--- a/service-worker/README.md
+++ b/service-worker/README.md
@@ -35,3 +35,7 @@ a sample illustrating multiple `fetch` handlers, each of which intercepts a diff
 
 - [Custom offline page](https://googlechrome.github.io/samples/service-worker/custom-offline-page/index.html) -
 a sample showing how to display a custom "Sorry, you're offline." error page when a network request fails.
+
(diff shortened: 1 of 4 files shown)


jakearchibald:
Cool! See also
https://jakearchibald.github.io/isserviceworkerready/demos/claim/

jeffposnick:
I'm going to interpret @jakearchibald's "Cool!" as a :+1: and merge this shortly, unless anyone else objects.

PR #103, opened 2015-04-09
window.caches example

jeffposnick:
@addyosmani @gauntface @PaulKinlan (@inexorabletash and @kenjibaheux as an FYI.)

This is an example of using `window.caches` to interact with the Cache Storage API. There is an associated service worker that pre-caches some resources, and you can use the in-page controls to add (using `window.fetch()`) and remove cache entries.

You can view a live demo at https://rawgit.com/jeffposnick/samples/sw-window-caches/service-worker/window-caches/index.html using Chrome 43 or later.

### DIFF
--- a/service-worker/README.md
+++ b/service-worker/README.md
@@ -35,3 +35,6 @@ a sample illustrating multiple `fetch` handlers, each of which intercepts a diff
 
 - [Custom offline page](https://googlechrome.github.io/samples/service-worker/custom-offline-page/index.html) -
 a sample showing how to display a custom "Sorry, you're offline." error page when a network request fails.
+
+- [Using `window.caches`](https://googlechrome.github.io/samples/service-worker/window-caches/index.html) -
+a sample showing how `window.caches` provides access to the Cache Storage API.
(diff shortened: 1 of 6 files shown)


addyosmani:
High-level first pass LGTM. I'll do a more thorough review tomorrow. Nice work getting this written so quickly!

jeffposnick:
My imagination very much veers towards the service worker use cases :smile:

I'm not having much luck coming up with an idea for a compelling demo that doesn't involve a service worker at all, but I wouldn't be surprised if there was the potential for something. I'm totally open to (someone) creating an additional, non-SW `cache-storage-api` sample if there are ideas.

In the meantime, I think this demo shows off the main methods for interacting with `window.caches` (adding, deleting, and listing entries), so hopefully it has some use.

addyosmani:
Thanks for the comments! I think in that light it makes sense for us to initially land this demo for the blog post and we can explore more independent examples in the cache-then-network case over in #106

addyosmani:
Sample LGTM. We might want to get a second look over by @jakearchibald before we ask you to rebase.

jakearchibald:
LGTM!

addyosmani:
@jeffposnick can you rebase?

addyosmani:
:+1:

PR #104, opened 2015-04-10
update-insecure-requests CSP demo.

jeffposnick:
@PaulKinlan @addyosmani @mikewest

Here's a demo of using `<meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests">` to put the new CSP into effect, along with a simple image embed showing off the functionality.

You can view a live version at https://rawgit.com/jeffposnick/samples/csp-upgrade-insecure-requests/csp-upgrade-insecure-requests/index.html

### DIFF
--- a/csp-upgrade-insecure-requests/index.html
+++ b/csp-upgrade-insecure-requests/index.html
@@ -47,6 +47,13 @@ limitations under the License.
 
     <link rel="icon" href="../images/favicon.ico">
 
+    <script>
+      // Ensure that this page is loaded via https:, since this demo won't be meaningful otherwise.
+      if ((!location.port || location.port === '80') && location.protocol !== 'https:') {
+        location.protocol = 'https:';
+      }
+    </script>
+
     <link rel="stylesheet" href="../styles/main.css">
   </head>
 
@@ -75,7 +82,7 @@ limitations under the License.
       <a href="http://www.html5rocks.com/en/tutorials/security/content-security-policy/#the-meta-tag">alternative</a>
       is to include the
       <code>&lt;meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests"&gt;</code>
-      tag in your HTML, preferably in the <code>&lt;head&gt;</code>.
+      tag in your HTML's <code>&lt;head&gt;</code>.
     </p>
 
     <div class="output">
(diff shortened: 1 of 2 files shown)


csp-upgrade-insecure-requests/index.html
```diff
@@ -0,0 +1,102 @@
...
+    </p>
+
+    <p>
+      Like other Content Security Policies, the recommend approach is to enable it via a HTTP
+      response header, <code>Content-Security-Policy: upgrade-insecure-requests</code>. However,
+      if you do not have control over the underlying web server (as is the case in this demo), an
+      <a href="http://www.html5rocks.com/en/tutorials/security/content-security-policy/#the-meta-tag">alternative</a>
+      is to include the
+      <code>&lt;meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests"&gt;</code>
+      tag in your HTML, preferably in the <code>&lt;head&gt;</code>.
```
mikewest: It has to be inside `<head>`, we'll ignore the policy if it appears in the body.

(Not sure you need to go into the detail, but you can inject it into the head via JavaScript; all that matters is that `<head>` is an ancestor of `<meta>`.)

csp-upgrade-insecure-requests/index.html
```diff
@@ -0,0 +1,102 @@
...
+    </p>
+
+    <div class="output">
+      <p>
+        The following image is loaded with an explicit <code>http:</code> URL,
+        <code>http://googlechrome.github.io/samples/images/touch/chrome-touch-icon-192x192.png</code>.
+        Because this page has <code>Content-Security-Policy: upgrade-insecure-requests</code>
+        active, the <code>http:</code> is treated as <code>https:</code>, and no mixed-content
+        warnings are displayed.
```
mikewest, line 94: Perhaps you could force the page onto HTTPS via JavaScript to make sure this is a true statement? :) (e.g. `if (window.location.protocol != "https:") { window.location.protocol = "https:"; }`)

jeffposnick, line 94: Makes sense. I put in the same redirect logic we're using for the service worker demos.

addyosmani:
Now that Mike's suggested changes are in, sample LGTM. Thanks Jeff!

PR #107, opened 2015-04-13
Add Permissions API sample

addyosmani:
r: @PaulKinlan @jeffposnick

### DIFF
--- a/permissions/index.html
+++ b/permissions/index.html
@@ -103,7 +103,7 @@ limitations under the License.
           updatePermission('geolocation', p.status);
 
           p.onchange = function() {
-            updatePermission('geolocation', p.status);
+            updatePermission('geolocation', this.status);
           };
         });
 
@@ -111,7 +111,7 @@ limitations under the License.
         navigator.permissions.query({name:'notifications'}).then(function(p) {
           updatePermission('notifications', p.status);
           p.onchange = function() {
-            updatePermission('notifications', p.status);
+            updatePermission('notifications', this.status);
           };
         });
 
@@ -119,14 +119,14 @@ limitations under the License.
         navigator.permissions.query({name:'midi', sysex:false}).then(function(p) {
           updatePermission('midi', p.status);
           p.onchange = function() {
-            updatePermission('midi', p.status);
+            updatePermission('midi', this.status);
           };
         });
 
         navigator.permissions.query({name:'midi', sysex:true}).then(function(p) {
           updatePermission('midi-sysex', p.status);
           p.onchange = function() {
-            updatePermission('midi-sysex', p.status);
+            updatePermission('midi-sysex', this.status);
           };
         });
 
--- a/permissions/index.html
+++ b/permissions/index.html
@@ -22,6 +22,7 @@ limitations under the License.
     <meta name="description" content="Sample illustrating the use of the Permissions aPI.">
 
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="manifest" href="manifest.json">
 
     <title>Permissions API Sample</title>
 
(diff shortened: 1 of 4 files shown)


permissions/index.html
```diff
@@ -0,0 +1,165 @@
...
+  <body>
+    <h1>Permissions API Sample</h1>
+
+    <p>Available in <a href="https://www.chromestatus.com/feature/6376494003650560">Chrome 43+</a></p>
+
+    <p>The Permissions API allows a web application to be aware of the status of a given permission, to know whether it is granted, denied or if the user will be asked whether the permission should be granted.</p>
+
+    <!-- // [START code-block] -->
+
+    <h3>Permissions</h3>
```
gauntface, line 56: Looking at the enum on the spec, the permission names are "geolocation", "notiications", "push" and "midi".

Is it just the case that push isn't implemented in Chrome yet?

https://w3c.github.io/permissions/#h-permission-registry

addyosmani, line 56: Latest details on what was implemented https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/exHjHQzY6m0, however I haven't been able to get push notifications working reliably with the API in Canary or ToT.

addyosmani, line 56: Got it working locally against push. Will try to add commits to expand the sample to cover.

permissions/index.html
```diff
@@ -0,0 +1,198 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+    <meta name="description" content="Sample illustrating the use of the Permissions aPI.">
```
jeffposnick: aPI :arrow_right: API

permissions/index.html
```diff
@@ -0,0 +1,198 @@
...
+            log('Geolocation permissions granted');
+            log('Latitude:' + position.coords.latitude);
+            log('Longitude:' + position.coords.longitude);
+          });
+        });
+
+        getNotBtn.addEventListener('click', function() {
+          Notification.requestPermission(function(result) {
+            if (result === 'denied') {
+              log('Permission wasn\'t granted. Allow a retry.');
```
jeffposnick: What does "Allow a retry." mean?

permissions/index.html
```diff
@@ -0,0 +1,198 @@
...
+          navigator.requestMIDIAccess({sysex:true});
+        });
+
+        getPushNotBtn.addEventListener('click', function() {
+          requestPushPermission();
+        });
+      }
+
+      init();
+    })(document);
```
jeffposnick: You're passing `document` in to the IIFE but not assigning it to a parameter, so I don't think there's any reason to pass it in.

permissions/index.html
```diff
@@ -0,0 +1,198 @@
...
+            updatePermission('notifications', p.status);
+          };
+        });
+
+        // Check for Midi/Midi SysEx permissions
+        navigator.permissions.query({name:'midi', sysex:false}).then(function(p) {
+          var update = function() { updatePermission('midi', p.status); };
+          update();
+
+          p.onchange = update;
```
jeffposnick: In a few of these you use the `p.onchange = update;` formulation, and in a few others you use

```
p.onchange = function() {
  updatePermission('notifications', p.status);
};
```

Is there a reason that some go with one vs. the other?

In general, what about using

```
p.onchange = function() {
  updatePermission('notifications', this.status);
};
```

for all of them, switching to using `this` instead of the closure over `p`?

jeffposnick:
Thanks for working on this!

Correct me if I'm wrong, but I don't believe you need any of the service worker bits. My understanding is that there's only one Notification-related permission (not a separate Push Notification one) and it doesn't look like you need to have a service worker registered in order to query its status or request access to it: http://jsbin.com/tajegozuyu/2/edit

There's no way that the service worker is going to trigger any notifications anyway (you need a real registration with a push server like GCM for that to happen), so adding in those bits just complicates the sample. The `manifest.json` can be left out, too.

Also, if @gauntface is going to write an accompanying updates.html5rocks.com article, then it may help to put in some of those code-snippet comments around the portions that could be pulled out and included inline.

gauntface:
I was the one who nudged the push permission - at the moment notification permission grants you push notification. This may not be the case in the future and I think this was largely to unblock push api from waiting on the permissions API.

Notifications and push are both referenced explicitly in the API doc.

gauntface:
Second look - yeah not sure if the SW bits are needed however - if it's a requirement for the push demo . . . . not sure what is best.

permissions/index.html
gauntface: +1 for this.status.

permissions/index.html
addyosmani: Good point. I should have dropped this earlier. Ditching in the next set of commits.

permissions/index.html
addyosmani: Dropping. There was some quirky behaviour during testing that suggested retries were possible, but I've confirmed in latest this isn't the case.

addyosmani:
Atm, Matt is correct in that both `push` and `notifications` can be queried based on the spec. 

For the sake of simplicity, I've dropped push notifications from the sample (as well as the SW code). I've also updated to address the comments on-thread.

PTAL.

gauntface:
LGTM

jeffposnick:
LGTM++.

I'm assuming the separate `push` permission is more forward thinking, since I've never run into a `push` permission prompt in Chrome 43. I can see it making sense as a standalone permission if/when push is decoupled from notifications, e.g. in the background sync use case.

addyosmani:
\o/

gauntface:
I think that is fair (ish) to say on regarding the permission.

PR #111, opened 2015-04-14
audiocontext.close sample

samthor:


addyosmani:
@cwilso will you be able to help review this? If not, we can but you have more Web Audio experience than we do :)

webaudio-audiocontext-close/index.html
```diff
@@ -0,0 +1,140 @@
...
+      try {
+        context = new AudioContext();
+      } catch (e) {
+        break;  // give up, can't create more contexts
+      }
+      pool.push(context);
+    }
+
+    function refreshCountMessage(opt_message) {
+      count.innerText = 'pool: ' + pool.length + ' AudioContext instances';
```
jeffposnick: `innerText` :arrow_right: `textContent` (https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#Differences_from_innerText)

webaudio-audiocontext-close/index.html
```diff
@@ -0,0 +1,140 @@
...
+      results.innerText = opt_message || '';
+    }
+    refreshCountMessage()
+
+    function closeRandomContext() {
+      if (pool.length == 0) {
+        refreshCountMessage();
```
jeffposnick: Is `refreshCountMessage();` necessary here? If it's not needed, then you could structure this as an `if (pool.length > 0) { ... }` and avoid the `return;`.

webaudio-audiocontext-close/index.html
```diff
@@ -0,0 +1,140 @@
...
+        return;
+      }
+      var choice = Math.floor(Math.random() * pool.length);
+      var context = pool[choice];
+
+      var promise = context.close();
+      promise.then(function() {
```
jeffposnick: `context.close().then(function() { ... })` is a bit more idiomatic. No need for the `promise` variable.

webaudio-audiocontext-close/index.html
```diff
@@ -0,0 +1,140 @@
...
+        pool.splice(pool.indexOf(context), 1);
+        refreshCountMessage('close promise resolved');
+      }, function(e) {
```
jeffposnick: Matter of personal preference, but (recently) instead of using the two-argument version of `then()`, I've found myself using a single-argument `then()` to handle the resolution followed by a `catch()` to handle the rejection.

Because it shares terminology with exceptions, developers who aren't familiar with promises will still get the idea as to what the `catch()` does, but seeing a two-argument `then()` doesn't give any hints.

webaudio-audiocontext-close/index.html
```diff
@@ -0,0 +1,140 @@
...
+        e = e || 'unknown';
+        refreshCountMessage('close promise failed to resolve: ' + e.toString());
```
jeffposnick: "failed to resolve" :arrow_right: "rejected"

webaudio-audiocontext-close/index.html
```diff
@@ -0,0 +1,140 @@
...
+      <button id="btn-create">Create</button>
+      <button id="btn-close">Close</button>
+      <pre id="results" style="color: red"></pre>
+    </div>
+
+    <!-- // [START code-block] -->
+    <script>
+
+    var pool = [];
+    var i = 0;
```
jeffposnick: Could you wrap lines 74-86 in an `init()` function or something similar? Might as well take the `i` and `context` variables out of the top-level scope.

jeffposnick:
Thanks, @samthor! I've added some comments that aren't very specific to the WebAudio API itself.
I'll defer to @cwilso as to whether it's showing off the right functionality, but unless he has an objections, it seems like the right approach to me.

cwilso:
From an audio functionality perspective, LGTM - tests the appropriate functionality.

webaudio-audiocontext-close/index.html
samthor: Done

webaudio-audiocontext-close/index.html
samthor: Done

webaudio-audiocontext-close/index.html
samthor: Agreed, I've taken this approach now.

webaudio-audiocontext-close/index.html
samthor: Done

webaudio-audiocontext-close/index.html
samthor: Done

webaudio-audiocontext-close/index.html
samthor: Done

samthor:
Reviewed and pushed, PTAL @jeffposnick

webaudio-audiocontext-close/index.html
```diff
@@ -0,0 +1,140 @@
...
+    }
+    refreshCountMessage()
+
+    function closeRandomContext() {
+      if (pool.length > 0) {
+        var choice = Math.floor(Math.random() * pool.length);
+        var context = pool[choice];
+
+        context.close().then(function() {
+          pool.splice(pool.indexOf(context), 1);
```
jeffposnick, line 105: I think you can replace `pool.indexOf(context)` with `choice` and it would be a bit clearer.

samthor, line 105: teeeechnically if the promise takes a while to resolve, you could be removing the wrong one. I've added a comment to that effect.

jeffposnick:
:+1:

PR #112, opened 2015-04-14
audionode disconnect sample, plus audio from h5r

samthor:
Demo for selective connect/disconnect of AudioNode.

Adds an audio file from h5r. I wonder whether it's worth moving out the (relatively small) `loadAudio` method into an audio helper, since it's a tiny bit of mostly irrelevant work. Probably fine for now though.


jeffposnick:
Do you think you'll use `techno.wav` elsewhere? If not, could you place it in the same directory as this sample? Our top-level directories are getting a bit overgrown as it is...

Also, just to double-check and because I haven't listened to it, is `techno.wav` a properly licensed piece of audio?

webaudio-audionode-disconnect/index.html
```diff
@@ -0,0 +1,134 @@
...
+    playerNode.loop = true;
+
+    document.getElementById('btn-play').addEventListener('click', function() {
+      this.disabled = true;
+      playerNode.start();
+    });
+
+    loadAudio("../audio/techno.wav").then(function(buffer) {
+      playerNode.source = buffer;
+    }, function(e) {
```
jeffposnick: See my comment in the other PR about using `catch()` instead of the two-argument `then()`, since it's arguably clearer to developers who don't know about promises.

jeffposnick:
I'm not seeing any logged errors, but I don't hear anything while trying out your demo locally. I see the `fetch()` complete, and I can click "Play", which is then disabled, but am not having much luck getting any sound. This is on Chrome Canary (M44), but presumably things launched in M43 weren't broken on M44.

Am I missing something? Can you double-check that it's functional?

webaudio-audionode-disconnect/index.html
```diff
@@ -0,0 +1,134 @@
...
+      alert('Could not load demo audio: ' + e.toString());
```
jeffposnick: Can you log this within `<div.output>` instead of using `alert()`?

samthor:
The fetch won't succeed unless you're running this on localhost - just
double-check the console for that.

The audio file is from www.htm5rocks.com's repo, which is Apache2, and
Google owned. I can move it back into this folder if you like - I just have
this problem that every WebAudio demo I write ends up just generating
tones, which are fine, but probably not very much fun to listen to.

jeffposnick:
Thanks for the confirmation re: the audio file's origins. Keeping it top-level sounds (hah!) good to me if you're going to use it elsewhere.

Regarding the audio in the demo, I'm running a local webserver with a clone of your repo and things look like they succeeded without any errors, but I'm not hearing anything. Here's what I see in DevTools after clicking Play:

image

webaudio-audionode-disconnect/index.html
samthor: Done

webaudio-audionode-disconnect/index.html
samthor: Done

samthor:
Sorry, I made a typo when refactoring the demo. It should play now (I was setting `.source`, not `.buffer`).

webaudio-audionode-disconnect/index.html
```diff
@@ -0,0 +1,136 @@
...
+
+    // mergerNode merges arbitrary audio into L/R channels.
+    var mergerNode = context.createChannelMerger(2);
+    mergerNode.connect(context.destination);
+    splitterNode.connect(mergerNode, 0, 1); // flip left output to rightinput
+
+    var connected = false;
+    document.getElementById('btn-connect').addEventListener('click', function() {
+      // Connect/disconnect right output to left input.
+      if (connected) {
```
jeffposnick, line 115: Can you change the text in the button depending on the `connected` state? I think it would help users confirm that the demo is working if they saw something update each time they clicked. (The audio effect is kind of subtle, at least on my laptop speakers.)

Assigning to `this.textContent` might be sufficient, since you're already inside an event listener for the button.

samthor, line 115: Done, and I've changed the wording to 'connect' and 'disconnect' to make it clearer what's going on.

jeffposnick:
:+1:

PR #113, opened 2015-04-14
Adding cut and copy sample

gauntface:
This is a basic sample of copy and paste. @addyosmani @jeffposnick the gate keepers of the sample world please advise.


gauntface:
Ill switch over to using log().

I would have thought the best feature detection would be queryCommandSupported(). The problem with this is that chrome is currently only returning true AFTER user interaction, at which point its too late to be useful.

Firefox says it's supported but throws an error when you try to use it.

Regarding ID's over classes - I just use classes after talking with Harry (a.k.a csswizardry). Have no strong opinions either way.

addyosmani:
LGTM

PR #120, opened 2015-04-20
s/Resource Requests/Requests/

mikewest:


mikewest:
Er. I'm a member of the Google group and everything... Also, I work there. :)

jeffposnick:
We love googlebot, but it's a little overzealous sometimes. Thanks, @mikewest

PR #123, opened 2015-04-29
Fix for mixed content error

karolklp:
It was working fine on localhost, but I got error on remote server:

> Mixed Content: The page at 'https://googlechrome.github.io/samples/service-worker/offline-analytics/index.html' was loaded over HTTPS, but requested an insecure image 'http://www.chromium.org/_/rsrc/1302286216006/config/customLogo.gif'. This content should also be served over HTTPS.

### DIFF
--- a/service-worker/offline-analytics/index.html
+++ b/service-worker/offline-analytics/index.html
@@ -93,8 +93,8 @@ limitations under the License.
       <div id="status"></div>
 
       <ul id="images" style="display: none">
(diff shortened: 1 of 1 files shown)


PaulKinlan:
Sorry, the point of this sample is to show the upgrade of resoruces to https (even if specified with http).  You need Chrome 43 (Current beta) to see this work.

PaulKinlan:
Sorry - Ignore me, was looking at something else.

jeffposnick:
:+1: Thanks!

PR #130, opened 2015-05-19
Extended Unicode escapes example.

jeffposnick:
R: @addyosmani @surma CC: @rossberg

Hopefully straightforward example of a string literal with an extended Unicode escape. There's apparently not very much else new to illustrate.

There's a live demo at https://rawgit.com/jeffposnick/samples/extended-unicode-escapes/extended-unicode-escapes/index.html

### DIFF
--- a/extended-unicode-escapes/README.md
+++ b/extended-unicode-escapes/README.md
@@ -1,0 +1,3 @@
+Extended Unicode Escapes in Strings Sample
+===
+See https://googlechrome.github.io/samples/extended-unicode-escapes/index.html for a live demo.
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
You might want to add a reference to the new ES6 `String.fromCodePoint` which returns a string created by using the specified sequence of code points.

jeffposnick:
It looks like `String.fromCodePoint()` snuck in under the radar into Chrome 41, and we didn't bother tracking that then. I can add in a paragraph about it here, I guess.

addyosmani:
Reviewed. Sample LGTM.

PR #137, opened 2015-06-25
Fix notificationonclick handler

jpmedley:
The test client.url == '/' doesn't work in every instance. I'll be happy to put something else if anyone has an idea.

### DIFF
--- a/push-messaging-and-notifications/service-worker.js
+++ b/push-messaging-and-notifications/service-worker.js
@@ -31,7 +31,7 @@ self.addEventListener('notificationclick', function(event) {
   }).then(function(clientList) {
     for (var i = 0; i < clientList.length; i++) {
       var client = clientList[i];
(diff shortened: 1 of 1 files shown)


gauntface:
I think the idea with that check is to highlight that when the notification is clicked, you need to only focus on a page that your notification would normally click to.

For example google.com could have a service worker for everything and send a notification for a new article and a notification for a G+ post. Without a check for the URL, when a notification is clicked for the article or G+ post, the click handler will find a client (http://google.com) and then just focus on that, even though the click should probably go to http://plus.google.com or http://google.com/news.

I'm happy to either leave this in as good practice or remove this all together and leave it opening a new tab everytime.

jpmedley:
I did some more experimenting and found a few other things about this focus issue. I'm closing this for now and will send an email around when I have some time.

PR #138, opened 2015-07-06
Adds performance test for React vs Vanilla

paullewis:
@PaulKinlan PTAL


paullewis:
@googlebot I signed it. On account of being a Googler and all that...

Issue #37, opened 2014-10-22
googlechrome.github.io needs landing page


ebidel:
That lists all the samples.

gauntface:
+1 this.

Maybe we could list samples and their relative chrome status page as well.

jeffposnick:
@gauntface, if you've got the interest to take this on either here or over at https://github.com/GoogleChrome/chromium-dashboard, I think that strictly speaking, anything would be better than the nothing that we currently have. I don't believe anyone else is currently looking into this.

PaulKinlan:
I played with it a while ago but we needed to integrate Jekyll more. If
someone is up for you will get much love frome me

P

jeffposnick:
After thinking about this some more, I'm of the opinion that this should be addressed via a new https://chromestatus.com/samples landing page, and we should create a top-level `index.html` in this repo that just contains a `<meta>` redirect there.

This repo is great for hosting new samples and offers a lot of flexibility around directory structure, including supporting JS/CSS files, and using HTTPS. This repo is not ideal for storing metadata about what each sample relates to (Networking, Multimedia, ES6, etc.) and metadata about which version of Chrome it's associated with. While we could try to add that metadata to each sample in a way that would be parseable by an index-generator, we'd effectively be replicating what already exists in the chromestatus.com backend. Without that metadata, putting together a meaningful browsing/search experience seems like an uphill battle.

Using a hypothetical https://chromestatus.com/samples as the canonical samples list also means that samples that predate this repo or that, for whatever reason, aren't hosted in this repo will show up in the list.

PR #140, opened 2015-07-07
Optional use of Jekyll templates for samples

### LINKED ISSUE #84: Templatize as much as possible
There's an enormous amount of boilerplate HTML (and some JS + CSS) that used in every sample. I'm guilty of encouraging a copy/paste model that got us to this point.

Moving to a model in which we took advantage of [GitHub's Jekyll support](https://help.github.com/articles/using-jekyll-with-pages/) and abstracted away most of the boilerplate into a shared template would make a lot of sense. Samples that needed to do something that didn't fit into the template would still have the flexibility to commit a full set of HTML + JS + CSS.

I think it would also be possible to automate generating an index page as part of this process, closing #37

jeffposnick:
R: @addyosmani @ebidel @PaulKinlan all

This modifies the recommended starting template to use a Jekyll-based template which takes care of most of the boilerplate.

Existing samples can stay as-is, since Jekyll will serve any page without frontmatter verbatim. (At some point, it would be nice to go back and update the ones that can be migrated easily.)

I've migrated a couple of samples over to the new template-based approach to illustrate the process:
- https://github.com/jeffposnick/samples/blob/gh-pages/report-validity/index.html ([live version](https://jeffy.info/samples/report-validity/index.html))
- https://github.com/jeffposnick/samples/blob/gh-pages/image-rendering-pixelated/index.html ([live version](https://jeffy.info/samples/image-rendering-pixelated/index.html))

One of the benefits of using the new templates is that we can easily show the relevant CSS/JS/HTML inline in the page with syntax highlighting in addition to using it in the source, which you can see in action in the live versions linked to above.

This also simplifies the master styles, removing some that were never used.

This closes #84


gauntface:
Is there any plan to add an index page for this?

jeffposnick:
The index page is being tracked in #37 and this PR doesn't address that.

I'm actually not sure whether putting together a full index page here makes the most sense vs. creating a new samples-focused page hosted on chromestatus.com. The advantage of hosting it on chromestatus.com is that it could use of all the metadata that we have about each feature (the category, the chrome release version, etc.) and of the App Engine datastore's ability to filter and search for entries. That might lead to a richer user experience than anything static we could host here.

Anyway, we can discuss more in #37 and potentially in a new issue in https://github.com/GoogleChrome/chromium-dashboard

SAMPLE_STARTING_POINT/index.html
```diff
@@ -0,0 +1,32 @@
+---
+feature_name: REPLACE_ME
+chrome_version: REPLACE_ME
+feature_id: REPLACE_ME
+local_css_files:
```
ebidel: can you give examples of these? is it a list of relative paths?

SAMPLE_STARTING_POINT/index.html
```diff
@@ -0,0 +1,32 @@
...
+local_js_files:
+additional_head_content:
```
ebidel: example

SAMPLE_STARTING_POINT/index.html
```diff
@@ -0,0 +1,32 @@
...
+
+{% capture initial_output_content %}
+<p>This can be used to set initial sample output.</p>
+{% endcapture %}
+{% include output_helper.html initial_output_content=initial_output_content %}
+
+{% capture js %}
+/*
+  Use this section for any JavaScript that shows off the feature's functionality.
+  Do not surround it with <script> tags.
```
ebidel: code block: `&lt;script>`

_includes/https_redirect.html
```diff
@@ -0,0 +1,8 @@
+<script>
+  // If we're running on a real web server (as opposed to localhost on a custom port,
+  // which is whitelisted), then change the protocol to HTTPS.
+  // See https://goo.gl/lq4gCo for an explanation.
+  if ((!location.port || location.port == "80") && location.protocol != 'https:') {
```
ebidel, line 5: can we just check `location.origin !== 'localhost'` so people can use their own port?

ebidel:
Reviewed

_includes/https_redirect.html
jeffposnick, line 5: I borrowed this logic from https://github.com/jakearchibald/isserviceworkerready/blob/master/src/base.html#L8

I think the idea is that checking for `location.origin !== 'localhost'` is overly restrictive, since there are IP4/IP6 addresses that are also whitelisted and would work.

The existing logic is a little hard to follow, but it does allow for use of localhost with any port other than 80.

Issue #150, opened 2015-07-14
Beacon Sample errors


PaulKinlan:
http://googlechrome.github.io/samples/beacon/

Uncaught SyntaxError: Unexpected token ( line 77

jeffposnick:
Closed in https://github.com/GoogleChrome/samples/pull/154

PR #154, opened 2015-07-15
Fixed beacon code sample

beaufortfrancois:
This patch fixes beacon code sample with missing parentheses.

cc @PaulKinlan 

BUG=#150


PaulKinlan:
LGTM. Not sure if we need another reviewer.

jeffposnick:
:+1:

PR #157, opened 2015-07-15
Center html body by default on desktop

beaufortfrancois:
This patch makes desktop samples more readable in my opinion. Hopefully you'll agree with that.

@jeffposnick


jeffposnick:
:+1: :straight_ruler:

PR #158, opened 2015-07-15
Set output and highlight style display to block

beaufortfrancois:
I think `inline-block` should not be the intended behaviour by default for `output` and `highlight`. This patch addresses this issue. 

@jeffposnick

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -16,7 +16,7 @@ h3 {
 .output {
   background-color: #f0f0f0;
   border-radius: 0.75em;
-  display: inline-block;
+  display: block;
   margin: 0.5em;
   padding: 0.5em;
 }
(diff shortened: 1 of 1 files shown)


jeffposnick:
The motivation for `inline-block` was that the default width would then be set to the actual content's width. When using `block`, the default width would be the full page, so there would be potentially a lot of empty space on the right before you get to the right-hand box border.

beaufortfrancois:
http://www.html5rocks.com/en/tutorials/service-worker/introduction/#toc-how and https://developers.google.com/web/updates/2015/06/mse-gapless-audio both use `block` which does make sense in my opinion.

jeffposnick:
Okay, in conjunction with https://github.com/GoogleChrome/samples/pull/157 this looks :+1: to me. Without that other PR, the boxes would be too big on desktop windows.

PR #160, opened 2015-07-15
Revert "Rewrite of the SW post message example"

jeffposnick:
Reverts GoogleChrome/samples#159


jeffposnick:
Yikes, that was supported to be merged into my fork, not the main repo. Reverted.

PR #161, opened 2015-07-15
Rewrite of the SW post message example

jeffposnick:
R: @addyosmani @wibblymat
CC: @mattto @jakearchibald

This is a rewrite of the existing sample showing off how to use `postMessage()` to communicate between the service worker and a controlled page.

The biggest change is that it shows off the new `Client.postMessage()` and associated `navigator.serviceWorker.onmessage` handler, for messages initiated from the service worker, in addition to showing messages going in the other direction.

The original sample was written in the M40 timeframe, so there are a few things that have changed since then in general: `Cache.add()` doesn't need a polyfill anymore, and `clients.claim()` is a thing.

It moves to the new templates, so a lot of the HTML boilerplate is gone.

You can view a live example (best in M45) at https://jeffy.info/samples/service-worker/post-message/

### DIFF
--- a/service-worker/post-message/index.html
+++ b/service-worker/post-message/index.html
@@ -1,207 +1,136 @@
-<!doctype html>
-<!--
-Copyright 2014 Google Inc. All Rights Reserved.
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


wibblymat:
LGTM

wibblymat:
I'll let @addyosmani give his LGTM too, in case I missed something :)

addyosmani:
LGTM2!

Issue #144, opened 2015-07-11
ES6 Array methods sample


jeffposnick:
See https://www.chromestatus.com/feature/6732923508097024

addyosmani:
@beaufortfrancois is going to take this one on

PR #163, opened 2015-07-16
Add Array Methods (ES6) Sample

beaufortfrancois:
Can someone review this Array Methods (ES6) Sample?

screenshot 2015-07-16 at 12 40 03 pm

BUG=#144

### DIFF
--- a/array-methods-es6/index.html
+++ b/array-methods-es6/index.html
@@ -1,0 +1,112 @@
+---
+feature_name: Array Methods (ECMAScript 2015)
+chrome_version: 45
+feature_id: 6732923508097024
+---
+<p><a target="_blank" href="http://www.ecma-international.org/ecma-262/6.0">ECMAScript 2015 specification</a> adds additional static methods on <code>Array</code> and instance
+methods on <code>Array.prototype</code>.</p>
+
+<h2>Array.from ( items [ , mapfn [ , thisArg ] ] ) <a target="_blank" href="http://www.ecma-international.org/ecma-262/6.0/#sec-array.from">#</a></h2>
+<p>
+The <code>Array.from()</code> method converts an array-like object to a true
+array. It takes an optional second parameter, which can be used to execute a map
+function on each element of the array that is being created.
+</p>
+{% capture array_from_js1 %}
+Array.from(new Set([1, 1, 2, 3]));         // [1, 2, 3]
+Array.from('hello');                       // ["h", "e", "l", "l", "o"]
+Array.from('hello', x => x.charCodeAt(0)); // [104, 101, 108, 108, 111]
+{% endcapture %}
+{% include js_snippet.html js=array_from_js1 title='' %}
+<p>
+It comes in handy when dealing with the infamous <code>NodeList</code>
+returned by <code>document.querySelectorAll</code>.
+</p>
+{% capture array_from_js2 %}
+var divs = document.querySelectorAll("div");
+Array.from(divs).forEach(function(node) {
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
I would be really happy if @littledan had some time to have a look at this and let me know if I've missed some somehow.

littledan:
This looks good to me.

array-methods-es6/index.html
```diff
@@ -0,0 +1,114 @@
+---
+feature_name: Array Methods (ES6)
+chrome_version: 45
+feature_id: 6732923508097024
+---
+<style>h2 { margin-top: 48px; }</style>
```
jeffposnick: Could you file a PR against the common CSS file to introduce this tweak if you feel like it significantly improves the layout? It doesn't make as much sense to be a one-off inline style.

jeffposnick:
Structural comment aside from the sample's content (which look good to me): I think it would be useful to show the output of the code, in addition to the code snippets. You're one of the first people writing samples after I added the new templates, so I'm sure the process isn't as clear as it could be. You could do something like https://github.com/GoogleChrome/samples/blob/gh-pages/report-validity/index.html#L25 and make use of the `ChromeSamples.*` helper methods to display the output of the JS methods in the "output helper" area.

array-methods-es6/index.html
beaufortfrancois: Thank you! 
See https://github.com/GoogleChrome/samples/pull/164

beaufortfrancois:
I've looked at the other es6 samples and I'm not sure there that displaying only the output of the JS methods in the "output helper" area would help people understand what is happening.

If we're talking about adding something like: 

``` html
<p>Array.from( <input type="text"> )</p>
<pre></pre>
```

And doing a simply evil eval on this, I think people should use Dev Tools instead.

The way I'm reading this sample is with my devtools opened and copy/playing commands on it. Hopefully I'm not the only one ;)

addyosmani:
Historically, our flow here (which could definitely be improved) has seen trying to display output on screen, but allowing the developer to otherwise dive into the source themselves and copy/paste to DevTools Console if they wanted to play around with the code a little further.

We don't really encourage opening up DevTools to preview output, but I personally wouldn't mind us doing this as we're otherwise replicating the console output in a limited form for the browser window. I'd be interested in whether we can figure out a nice balance here, but we might want to table that discussion until after the M45 samples land so we have more time to think it through properly (will probably affect existing samples too).

Deferring final call on this to @jeffposnick

jeffposnick:
So what I was thinking was something along the lines of:

```
{% include output_helper.html initial_output_content=initial_output_content %}

{% capture array_from_js1 %}
var arrayFromSet = Array.from(new Set([1, 1, 2, 3]));
ChromeSamples.log('arrayFromSet: ', arrayFromSet);
var arrayFromString = Array.from('hello');
ChromeSamples.log('arrayFromString: ', arrayFromString);
var arrayFromStringWithMap = Array.from('hello', x => x.charCodeAt(0));
ChromeSamples.log('arrayFromStringWithMap: ', arrayFromStringWithMap);
{% endcapture %}
{% include js_snippet.html js=array_from_js1 %}
```

But honestly the output isn't great there, either, and wouldn't work well with multiple JS snippets.

Rather than having you restructure everything right now, I'm fine with merging this, and I've filed https://github.com/GoogleChrome/samples/issues/167 to track coming up with a cleaner solution to this pattern.

addyosmani:
SGTM.

PR #156, opened 2015-07-15
Added customizable snippet title to _include

beaufortfrancois:
While I was working on Array Methods (ES6) Sample, I needed to customize title for JS snippets. Hopefully it'll get approved.

```

{% include js_snippet.html js=js %}
<h3>Javascript Snippet</h3>

{% include js_snippet.html js=js title='Snippet' %}
<h3>Snippet</h3>

{% include js_snippet.html js=js title='' %}
<EMPTY>
```

@jeffposnick

PR #166, opened 2015-07-20
Swap around if/else logic around titles in templates

jeffposnick:
R: @beaufortfrancois 

I think the logic in https://github.com/GoogleChrome/samples/pull/156 needs to be tweaked a bit. I noticed that leaving out the snippet title parameter was leading to an empty `<h3>` as opposed to the default `<h3>`.

Compare https://jeffy.info/samples/report-validity/ (with this PR) to http://googlechrome.github.io/samples/report-validity/ (without this PR).

I'm also adding some indentation to make the logic clearer.

### DIFF
--- a/_includes/css_snippet.html
+++ b/_includes/css_snippet.html
@@ -1,11 +1,13 @@
 {% if include.css %}
-<style>{{ include.css }}</style>
-{% if include.title != "" %}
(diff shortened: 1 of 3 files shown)


beaufortfrancois:
Sorry for that.
You're absolutely right.

LGTM

PR #169, opened 2015-07-21
Add Object.assign() sample (fixes #152)

### LINKED ISSUE #152: Object.assign sample
> Object.assign() now makes it easy for developers to clone or merge objects by copying all the values of enumerable properties from one or more source objects to a target object.

addyosmani:
r: @jeffposnick 

Minor feedback on templating: I found myself needing to `JSON.stringify()` object output in order for it to display correctly in the ChromeSamples output area. To make it easier for someone to take the source and modify it for use in the console without this, we may want to make `ChromeSamples.log` a little smarter when it comes to type detection, but this may also be out of scope.

### DIFF
--- a/object-assign-es6/index.html
+++ b/object-assign-es6/index.html
@@ -1,0 +1,46 @@
+---
+feature_name: ES6 Object.assign()
+chrome_version: 45
+feature_id: REPLACE_ME
+additional_head_content: <meta name="optional" value="optional">
+---
+
+<h3>Background</h3>
+<p><code>Object.assign()</code> copies the values (of all enumerable own properties) from one or more source objects to a target object. It is useful for merging objects or cloning them shallowly.</p>
+
+{% capture initial_output_content %}
+{% endcapture %}
+{% include output_helper.html initial_output_content=initial_output_content %}
+
+{% capture js %}
+'use strict';
+
+// Merge an object
+let first = { name: "Tony" };
+let last = { lastName: "Stark" };
+let person = Object.assign(first, last);
+ChromeSamples.log(JSON.stringify(person));
+// {name: "Tony", lastName: "Stark"}
+ChromeSamples.log(JSON.stringify(first));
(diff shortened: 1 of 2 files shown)


object-assign-es6/index.html
```diff
@@ -0,0 +1,46 @@
+---
+feature_name: ES6 Object.assign()
+chrome_version: 45
+feature_id: REPLACE_ME
+additional_head_content: <meta name="optional" value="optional">
```
jeffposnick: You can leave this out.

object-assign-es6/index.html
```diff
@@ -0,0 +1,46 @@
...
+// Clone an object
+let obj = { person: "Thor Odinson"};
+let clone = Object.assign({}, obj);
+ChromeSamples.log(JSON.stringify(clone));
+// {person: "Thor Odinson"}
+
+{% endcapture %}
+{% include js_snippet.html js=js %}
+
+{% include css_snippet.html css=css %}
```
jeffposnick: You can leave this out.

object-assign-es6/index.html
```diff
@@ -0,0 +1,46 @@
+---
+feature_name: ES6 Object.assign()
+chrome_version: 45
+feature_id: REPLACE_ME
+additional_head_content: <meta name="optional" value="optional">
+---
+
+<h3>Background</h3>
+<p><code>Object.assign()</code> copies the values (of all enumerable own properties) from one or more source objects to a target object. It is useful for merging objects or cloning them shallowly.</p>
```
jeffposnick: Can you put in something about how the target object is the first parameter, and that the target object is also used as the return value? The feature is somewhat nonintuitive in that regard, and I could imagine folks getting confused.

jeffposnick:
If you merge #174 first, you can leave out the `JSON.stringify()` calls from this sample.

addyosmani:
Feedback addressed. PTAL if you get a chance.

jeffposnick:
:+1:

PR #173, opened 2015-07-21
Top-level index.html redirect to chromestatus.com/features

jeffposnick:
R: @PaulKinlan @gauntface @addyosmani @ebidel

@ebidel is going to be putting together a real samples landing page (see GoogleChrome/chromium-dashboard#216)

In the meantime, re: #37, here's a redirect to the main chromestatus.com features list. Once the samples landing page is created, we can update the redirect to point there instead.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,7 @@
+<!doctype html>
+<html>
+<head>
+  <title>Redirecting...</title>
+  <meta http-equiv="refresh" content="0;URL='https://www.chromestatus.com/features'">
+</head>
+</html>


beaufortfrancois:
qq: Why not https://help.github.com/articles/redirects-on-github-pages?

jeffposnick:
Partly because I had no idea it existed :smile: 

But looking at the underlying code, it seems more straightforward to just explicitly write out the HTML. The Ruby gem doesn't do much extra and introduces another dependency: https://github.com/jekyll/jekyll-redirect-from/blob/master/lib/jekyll-redirect-from/redirect_page.rb#L21

ebidel:
LGTM.

PR #174, opened 2015-07-21
Call JSON.stringify() on non-string arguments while logging

jeffposnick:
R: @addyosmani 

Addresses the feedback you had in https://github.com/GoogleChrome/samples/pull/169 about logging.


addyosmani:
per #169, will merge this in before moving ahead.

PR #175, opened 2015-07-21
Reverts a few body styles

jeffposnick:
R: @beaufortfrancois 

I'm not sure how these extra styles ended up being applied to `body` (they got picked up in this merge: https://github.com/GoogleChrome/samples/commit/d1fc4d9fadaef5fb65c56f50b88627cab9470f36), but I don't think they're intentional and removing them makes things look a bit better on smaller viewports.


beaufortfrancois:
LGTM except commit title.

PR #180, opened 2015-07-22
Snipping away a couple of extra boilerplate lines

jeffposnick:
R: @addyosmani 

Sigh, a couple of extra lines slipped through that need to be trimmed out.


beaufortfrancois:
LGTM

PR #183, opened 2015-07-23
Full Travis CI configuration

jeffposnick:
R: Any/all

This should be enough to get a Travis CI build going for the Jekyll content that's checked in to any branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,5 @@ branches:
   only:
   - /.*/
 cache: bundler
+notifications:
+  email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,5 @@
 language: ruby
 script: "bundle exec jekyll build"
+branches:
+  only:
+  - /.*/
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,2 +1,10 @@
 language: ruby
+rvm:
+- 2.1
 script: "bundle exec jekyll build"
```
PaulKinlan: I would change this to 
script:
- bundle exec jekyll build

it just helps for the future.

PaulKinlan: Also are we absolutely sure that jekyll is deployed on the travis infrastructure?  It seems to make sense to have an

install:
- bundle install

jeffposnick: Travis CI automatically runs that for you when it detects that there's a `Gemfile`: http://docs.travis-ci.com/user/languages/ruby/#Travis-CI-uses-Bundler

You can't use `cache: bundler` if you override the install step: http://docs.travis-ci.com/user/caching/#With-a-custom-install-step, so it's cleaner to leave things as-is and take advantage of the default behavior.

jeffposnick:
I'm going to merge, having made the change @PaulKinlan suggested.

Issue #141, opened 2015-07-07
Use eslint to standardize on JS code style


jeffposnick:
It's been suggested in the past (by @gauntface I believe) to include a top-level [`jscs`](http://jscs.info/) configuration in this repo. It might be difficult to automatically enforce the styles, but including a reference that folks could opt into would do some good.

addyosmani:
:+1:

jeffposnick:
`jscs` doesn't currently support extracting JS `<script>` blocks from HTML files (at least not until [this PR](https://github.com/jscs-dev/node-jscs/pull/1038) is merged).

`jshint` appears to do a decent job of extracting out JS `<script>` tags, though. While `jshint`'s focus is slightly different that `jscs`'s, it's still useful, and I'm going to look into use it instead.

jeffposnick:
I think this can be handled by moving to `eslint` with a standard config.

gauntface:
Now that everything is getting externalised I think we could do this fairly
soon when PRs calm down for templatizing.

jeffposnick:
Yup, though since it makes sense to use `eslint` for code style, we'll need a reference set of rules to use as a base. I believe @addyosmani is looking into that.

PR #184, opened 2015-07-23
jshint-based linting

jeffposnick:
R: @addyosmani @PaulKinlan @gauntface _et alii_

This introduces a new `linter.rb` script, which can be run either manually or via `.git/hook/pre-commit` (recommended), and checks for the following:
- a successful Jekyll site build
- for each (committed) file, that all boilerplate content has been removed
- for each (committed) file, that the templated version of it under _site/ passes `jshint`

The idea is that for now this would be recommended for all new/modified files, and we'd also clean up the backlog/adjust `.jshintrc` to make `jshint` happy with everything. At that point, we can update Travis CI to automatically lint the full site as part of every PR.

Related to https://github.com/GoogleChrome/samples/issues/141, but not enough to close it out yet.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,13 @@
+{
+  "esnext": true,
+  "expr": true,
+  "globalstrict": true,
+  "strict": false,
+  "globals": {
+    "console": false,
+    "self": false,
+    "window": false,
+    "navigator": false,
+    "document": false
+  }
+}
--- a/README.md
+++ b/README.md
@@ -28,3 +28,20 @@ the functionality. The email address of the engineer who worked on a given featu
 the corresponding https://www.chromestatus.com/features entry. If you're unsure of the GitHub
 username corresponding to the engineer, an alternative is to email them a link to the pull request
 and ask for feedback directly.
+
+Style / Linting / CI
+===
+The samples ideally should follow the [Google JavaScript Style Guide](http://google.github.io/styleguide/javascriptguide.xml),
+though this is not automatically enforced. (See https://github.com/GoogleChrome/samples/issues/141)
+
+It's recommended that all new samples pass the [`jshint`](http://jshint.com/install/) linting
+process, using the customized [`.jshintrc`](.jshintrc) settings. To automate linting against
+JavaScript in the Jekyll-ized HTML output, [`linter.rb`](linter.rb) can be used. Regular
+contributors should set up a git pre-commit
+[hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) to run [`linter.rb`](linter.rb)
+against all files that are being committed, via `ln -s -f ../../linter.rb .git/hooks/pre-commit`
+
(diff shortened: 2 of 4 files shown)


PaulKinlan:
Some minor comments:
- Should we not use a [rake](https://github.com/ruby/rake) file and have a task for test/build that includes the linter.
- It feels rather than a commit hook we can get travis to build and test all PR's and reject them if they break the build.

jeffposnick:
Using [`rake`](https://github.com/ruby/rake) makes sense. I'll refactor this PR to do that.

Regarding the Travis CI integration, my thinking is that we should get all the pre-existing samples cleaned up first (which will need to be done in future PRs), and once we get to a "clean" baseline, start linting the entire generated site as part of the Travis CI checks. I don't know of a clean way to tell Travis CI to just check a subset of changed files rather than the whole generated site, while that's possible to do via a `pre-commit` hook.

addyosmani:
> Regarding the Travis CI integration, my thinking is that we should get all the pre-existing samples cleaned up first (which will need to be done in future PRs), and once we get to a "clean" baseline, start linting the entire generated site as part of the Travis CI checks. 

This sgtm.

jeffposnick:
PTAL for this stage in the continuing process of implementing continuous integration/linting.

I've addressed some of the comments, but stopped short of wrapping up everything in a `Rakefile` at this point. When I tried that approach, the `.git/hook/pre-commit` setup ended up being more complicated—it would need to be a wrapper shell script which in turn invoked `rake` which in turn executed the existing Ruby code.

In a follow-up PR that adds linting to the Travis CI build process I may wrap things up in a `Rakefile`, but I want to approach this in smaller steps.

gauntface:
I struggled with this tbh.

Having to run this on pre-commit only is frustrating when I'd rather know while I'm working on the code. When there is an issue in the pre-commit step it simply states there was an issue, not what the issue was or where it resides.

I ended up creating a gulptask that just watches for changes and runs everything through JSHint. This had its own issues in that it doesn't like the yaml at the top of the files using jekyll templates. To overcome this I had to jekyll serve, point gulp at the _site directory and then basically find the errors in the final html page, then match that back to the source and fix it there, presumably this would be the same for the pre-commit hook?

When it came to actually commiting the final code, the pre-commit hook just output:

```
.git/hooks/pre-commit:34:in ``': No such file or directory - bundle (Errno::ENOENT)
    from .git/hooks/pre-commit:34:in `<main>'
```

It had worked prior to this outputting:

```
.git/hooks/pre-commit:68:in `<main>': One or more errors were found. (RuntimeError)
```

Was there a reason these wasn't just moved to a build process where it could be used during development?

I have zero confidence if everything is fix, but I had a crack here: https://github.com/gauntface/samples

Happy to make a PR, but it's not validated again the linter.rb file so have held off for now.

jeffposnick:
I'm happy to defer to your approach, @gauntface.

https://www.npmjs.com/package/eslint-plugin-html might allow us to use `eslint` instead of `jshint`, which would be really nice.

PR #186, opened 2015-07-24
Redirect to HTTPS for all pages using the new templates.

jeffposnick:
R: @PaulKinlan @wibblymat _et alii_

It was previous opt-in for pages using the new templates (via `https_redirect: true`), but there's no reason not to enable it by default for everything on the templates.

(As far as I can tell, you can't opt-in to [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) for GitHub pages, but if anyone knows how, that would be a nice alternative. I'd be a little concerned that some of our older samples might have mixed-content issues if HSTS were possible, so that would require some scrubbing first.)

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -21,7 +21,14 @@ limitations under the License.
     <meta name="description" content="Sample illustrating the use of {{ page.feature_name }}.">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>{% if page.title %}{{ page.title }}{% else %}{{ page.feature_name }} Sample{% endif %}</title>
-    {% if page.https_redirect %}{% include https_redirect.html %}{% endif %}
+    <script>
+      // If we're running on a real web server (as opposed to localhost on a custom port,
+      // which is whitelisted), then change the protocol to HTTPS.
+      // See https://goo.gl/lq4gCo for an explanation as to why this is needed for some features.
+      if ((!location.port || location.port == "80") && location.protocol != 'https:') {
+        location.protocol = 'https:';
+      }
+    </script>
     {% assign sub_dirs = page.url | split: '/' | size | minus: 2 %}
     {% capture relative_path_to_root %}{% for i in (1..sub_dirs) %}../{% endfor %}{% endcapture %}
     <link rel="icon" href="{{ relative_path_to_root }}images/favicon.ico">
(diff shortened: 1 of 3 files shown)


PaulKinlan:
LGTM.

_layouts/default.html
```diff
@@ -21,7 +21,14 @@
     <meta name="description" content="Sample illustrating the use of {{ page.feature_name }}.">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>{% if page.title %}{{ page.title }}{% else %}{{ page.feature_name }} Sample{% endif %}</title>
-    {% if page.https_redirect %}{% include https_redirect.html %}{% endif %}
+    <script>
+      // If we're running on a real web server (as opposed to localhost on a custom port,
+      // which is whitelisted), then change the protocol to HTTPS.
+      // See https://goo.gl/lq4gCo for an explanation as to why this is needed for some features.
+      if ((!location.port || location.port == "80") && location.protocol != 'https:') {
```
beaufortfrancois, line 28: Would we prefer `location.protocol == 'http'` instead?

jeffposnick, line 28: A `location.protocol` of 'http' is okay for whitelisted localhost-like domains ('localhost', '127.0.0.1', some others listed at https://goo.gl/lq4gCo) so the more nuanced check is required.

beaufortfrancois: Sorry I meant `if ((!location.port || location.port == "80") && location.protocol == 'http:') {`

beaufortfrancois, line 28: I meant `if ((!location.port || location.port == "80") && location.protocol == 'http:') {`

jeffposnick, line 28: I'm following @jakearchibald's lead here, e.g.: https://github.com/jakearchibald/svgomg/blob/02f4dd3666aee701792420cdf52b8f7819fc05f8/src/index.html#L9

Are you suggesting this to avoid a redirect when the page is loaded with via `file:`? Or is there some other benefit?

beaufortfrancois, line 28: I'd argue it's best practise to perform a whitelist than a blacklist. 
In this case, we want to redirect to `https` only if protocol is `http`.
If it's `file:`, `ftp` or whatever "protocol", it shouldn't do a redirect.

I'm nitpicking there.
Feel free to ignore this and merge it ;)

jeffposnick, line 28: @PaulKinlan merged it 5 hours ago :smile:

beaufortfrancois, line 28: We're good then :yum:

PR #189, opened 2015-07-28
Added Block modal dialogs in a sandboxed iframe Sample

beaufortfrancois:
Hello @mikewest,

Can you review this sample for modal dialogs blocked in a sandboxed frame?
If you don't want to run the sample, here's a quick look at it:

screenshot 2015-07-28 at 12 12 05 pm

### DIFF
--- a/block-modal-dialogs-sandboxed-iframe/README.md
+++ b/block-modal-dialogs-sandboxed-iframe/README.md
@@ -1,0 +1,5 @@
+Block modal dialogs inside a sandboxed iframe Sample
+====================================================
+See https://googlechrome.github.io/samples/block-modal-dialogs-sandboxed-iframe/index.html for a live demo.
+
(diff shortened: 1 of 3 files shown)


beaufortfrancois:
cc @jeffposnick for reviewing general look.

mikewest:
The code LGTM, thanks for putting this together! Will you do something
similar for "allow-popups-to-escape-sandbox"?

beaufortfrancois:
I will ;)

jeffposnick:
The comments from https://github.com/GoogleChrome/samples/pull/191 apply here as well.

beaufortfrancois:
Thank you @jeffposnick for this valuable feedback. I didn't realize html_snippet was actually meant for that. I've addressed all your feedback there.

beaufortfrancois:
That makes the sample much better. See below
screenshot 2015-07-31 at 11 08 06 am

PR #190, opened 2015-07-28
Added `clients.claim()` to the notification vibrate sample

wibblymat:
The notification sample didn't work on first page load because it waited for `serviceWorker.ready` without having `clients.claim()` in the worker.

R: @jeffposnick @PaulKinlan


jeffposnick:
:+1:

PR #191, opened 2015-07-29
Added Allow popups to escape sandboxed iframe Sample

beaufortfrancois:
@mikewest Your wish is my command.

I could have used `srcdoc` but I wanted to focus on the `sandbox` attribute there. 
Hope it's ok with you.

cc @jeffposnick 

screenshot 2015-07-29 at 4 05 52 pm

### DIFF
--- a/allow-popups-to-escape-sandbox/index.html
+++ b/allow-popups-to-escape-sandbox/index.html
@@ -1,0 +1,50 @@
+---
+feature_name: Allow popups to escape sandboxed iframe
+chrome_version: 46
+feature_id: 5708368589094912
+---
+
+<p>In order to mitigate the risks associated with ads when using sandboxed
+iframes, developers can now add <code>"allow-popups-to-escape-sandbox"</code>
+to the frame <code>"attribute"</code>. This allows the sandboxed document to
+spawn new windows without forcing the sandboxing flags upon them, hence
+creating a clean browsing context. For instance, a third-party
+advertisement could be safely <b>sandboxed without forcing the same restrictions
+upon a landing page</b>.</p>
+
+<p>Popup windows can be spawned with <code>window.open()</code> and
+<code>target="_blank"</code> links.</p>
+
+{% capture html %}
+<!-- No sandbox there... Popup window won't be sandboxed as well -->
+<iframe id="red" src="iframe.html"></iframe>
+
+<!-- This sandboxed frame will allow sandboxed popup window to open popups
+     but not to execute JavaScript for instance. -->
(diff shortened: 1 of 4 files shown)


mikewest:
LGTM (though here and in the other CL, it's now not clear to me why you're building the markup with JavaScript rather than just doing `<iframe sandbox="...">` in markup)

beaufortfrancois:
I think it's easier to read/understand what is happening in each iframe by showing user JavaScript.

allow-popups-to-escape-sandbox/index.html
```diff
@@ -0,0 +1,68 @@
+---
+feature_name: Allow popups to escape sandboxed iframe
+chrome_version: 46
+feature_id: 5708368589094912
+---
+
+<p>Google's anti-malvertising team wants to mitigate the risks associated with
```
jeffposnick: I'd be concerned about including "Google's _XYZ_ team wants so-and-so" in the samples as justification for a new feature. This feature is going through the W3C process, so whatever the initial motivation was, the spec is going to end up getting feedback from other members of the wider open web community.

Could you rephrase as _"In order to mitigate the risks associated with ads when using sandboxed iframes, developers can now add..."_

jeffposnick:
I'm going to go with @mikewest here and suggest that this would make more sense if we stuck to using HTML to construct the DOM. To use an analogy, I'd find it odd if we were showing off some new CSS functionality with `setAttribute('style', 'abc: xyz');` rather than using a `<style>` element or stylesheet, which is what most developers would end up using.

You can use [`html_snippet.html`](https://github.com/GoogleChrome/samples/blob/gh-pages/_includes/html_snippet.html) to write HTML and include it verbatim on the page:

```
{% capture html %}
<!-- No sandbox there... Popup window won't be sandboxed as well. -->
<h4>No Sandbox</h4>
<iframe src="iframe.html"></iframe>

<!-- This sandboxed frame will allow sandboxed popup window to open popups but not to execute
     JavaScript for instance. -->
<h4>sandbox="allow-popups"</h4>
<iframe src="iframe.html" sandbox="allow-popups"></iframe>

<!-- This sandboxed frame will create a clean non sandboxed popup window, allowed to execute
     JavaScript and open popups. -->
<h4>sandbox="allow-popups allow-popups-to-escape-sandbox"</h4>
<iframe src="iframe.html" sandbox="allow-popups allow-popups-to-escape-sandbox"></iframe>
{% endcapture %}
{% include html_snippet.html html=html %}
```

beaufortfrancois:
Thanks again @jeffposnick!
All your feedback has been addressed.

screenshot 2015-07-31 at 11 06 05 am

PR #193, opened 2015-08-05
Added user bluetooth service name to device-info bt sample

beaufortfrancois:
Instead of having a hard-coded `generic_access` bluetooth service, let's give user the ability to scan for a nearby device of his choice.

R: @jeffposnick

### DIFF
--- a/web-bluetooth/device-info.html
+++ b/web-bluetooth/device-info.html
@@ -9,22 +9,63 @@ feature_id: 5264933985976320
 <p>This sample illustrates the use of the Web Bluetooth API to retrieve basic
 device information from a nearby Bluetooth Low Energy Device.</p>
 
-<button>Get Bluetooth Device Info</button>
+<form>
+  <label for="service">
+    Enter a Bluetooth Service name:
+    <input id="service" type="text" list="services" autofocus />
+  </label>
+  <datalist id="services">
+    <option value="alert_notification">
+    <option value="automation_io">
+    <option value="battery_service">
+    <option value="blood_pressure">
+    <option value="body_composition">
+    <option value="bond_management">
+    <option value="continuous_glucose_monitoring">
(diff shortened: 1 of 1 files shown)


web-bluetooth/device-info.html
```diff
@@ -9,22 +9,63 @@
 <p>This sample illustrates the use of the Web Bluetooth API to retrieve basic
 device information from a nearby Bluetooth Low Energy Device.</p>
 
-<button>Get Bluetooth Device Info</button>
+<form>
+  <label for="service">
+    Enter a Bluetooth Service name:
+    <input id="service" type="text" list="services" autofocus />
```
jeffposnick: You can get rid of the self-closing `/` at the end of the tag.

web-bluetooth/device-info.html
```diff
@@ -9,22 +9,63 @@
...
+  <button>Get Bluetooth Device Info</button>
+</form>
 {% capture initial_output_content %}
 {% endcapture %}
 {% include output_helper.html initial_output_content=initial_output_content %}
 
 {% capture js %}
-function onButtonClick() {
+function onFormSubmit() {
+  var filterService = document.getElementById('service').value.trim();
```
jeffposnick: Why's the `.trim()` needed? I wouldn't assume that users would enter whitespace around their service name.

web-bluetooth/device-info.html
beaufortfrancois: Done.

web-bluetooth/device-info.html
beaufortfrancois: Removed.

beaufortfrancois:
@jeffposnick Thanks for the review. 
I've addressed both nits.

jeffposnick:
:+1:

PR #196, opened 2015-08-14
Correct URL

jeffposnick:
R: @surma


surma:
The smaller the PR the more often something like this happens to me. I need to be more thorough. Thanks for keeping an eye out, @jeffposnick

PR #198, opened 2015-08-21
[web-bluetooth device-info] Handle short 16- and 32-bit form

beaufortfrancois:
This will be useful for developers who use a lot of custom service UUIDs like me ;) 

R= @jeffposnick


beaufortfrancois:
@jeffposnick (gentle ping)

jeffposnick:
I was going to say that maybe you could just unconditionally call `parseInt()`, and then I looked up what the [spec says](https://webbluetoothcg.github.io/web-bluetooth/#typedefdef-bluetoothserviceuuid) about that parameter—it can be a 16- or 32-bit integer, or a UUID string, or an assigned service string, or... some other type of string? That's a funky API!

So anyway, :+1:

PR #199, opened 2015-08-26
CSS motion path sample

jeffposnick:
R: @addyosmani @PaulKinlan _et al._

This ports @ericwilligers CSS motion path demo from http://codepen.io/ericwilligers/pen/pJarJO to a more permanent home in the Chrome Samples repo, following the format of the other samples.

You can see a live preview of the sample landing page at https://jeffy.info/samples/css-motion-path/ using Chrome 46.

### DIFF
--- a/css-motion-path/index.html
+++ b/css-motion-path/index.html
@@ -34,7 +34,7 @@ feature_id: 6190642178818048
         stroke-width="13"
         stroke-linejoin="round"
         stroke-linecap="round" />
-  <path id="first"
+  <path id="firstScissorHalf"
         d="M30,0 H-10 A10,10 0 0,0 -20,10 A20,20 0 1,1 -40,-10 H20 A10,10 0 0,1 30,0 M-40,20 A10,10 1 0,0 -40,0 A10,10 1 0,0 -40,20 M0,0"
         transform="translate(0,0)"
         fill="green"
@@ -43,7 +43,7 @@ feature_id: 6190642178818048
         stroke-linejoin="round"
         stroke-linecap="round"
         fill-rule="evenodd" />
-  <path id="second"
+  <path id="secondScissorHalf"
         d="M30,0 H-10 A10,10 0 0,1 -20,-10 A20,20 0 1,0 -40,10 H20 A10,10 0 0,0 30,0 M-40,-20 A10,10 1 0,0 -40,0 A10,10 1 0,0 -40,-20 M0,0"
         transform="translate(0,0)"
         fill="forestgreen"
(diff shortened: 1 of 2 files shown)


css-motion-path/index.html
```diff
@@ -0,0 +1,87 @@
...
+           fill="yellow" />
+  <polygon points="993,245 993,190 1086,190 1086,300"
+           fill="red" />
+  <path d="M900,190 L993,245 V201 A11,11 0 0,1 1004,190 H1075 A11,11 0 0,1 1086,201 V300 L1294,423 H1216 A11,11 0 0,0 1205,434 V789 A11,11 0 0,1 1194,800 H606 A11,11 0 0,1 595,789 V434 A11,11 0 0,0 584,423 H506 L900,190"
+        fill="none"
+        stroke="black"
+        stroke-width="13"
+        stroke-linejoin="round"
+        stroke-linecap="round" />
+  <path id="first"
```
beaufortfrancois: You might want to rename it to `firstScissor` and `secondScissor` so that it's clear to developers what we're animating.

css-motion-path/index.html
```diff
@@ -0,0 +1,87 @@
...
+
+{% capture css %}
+#first {
+  motion-path: path('M900,190  L993,245 V201  A11,11 0 0,1 1004,190  H1075  A11,11 0 0,1 1086,201  V300  L1294,423 H1216  A11,11 0 0,0 1205,434  V789  A11,11 0 0,1 1194,800  H606  A11,11 0 0,1 595,789  V434  A11,11 0 0,0 584,423  H506 L900,190');
+  transform: translate(0px,0px);
+}
+
+#second {
+  motion-path: path('M900,190  L993,245 V201  A11,11 0 0,1 1004,190  H1075  A11,11 0 0,1 1086,201  V300  L1294,423 H1216  A11,11 0 0,0 1205,434  V789  A11,11 0 0,1 1194,800  H606  A11,11 0 0,1 595,789  V434  A11,11 0 0,0 584,423  H506 L900,190');
+  transform: translate(0px,0px);
```
beaufortfrancois: I'm not sure `transform` is needed there.

css-motion-path/index.html
addyosmani: I agree with the rename if possible. It does make it more clear on first read.

jeffposnick: Done. I've gone with `firstScissorHalf` and `secondScissorHalf` to make it clear that each one is half of the full object.

css-motion-path/index.html
jeffposnick: It was in @ericwilligers's original example, but the demo still works the same (as far as I can tell) with it removed, so I'll leave it out.

jeffposnick:
Thanks! I'll merge this in a little bit unless there are any other points of feedback.

addyosmani:
🌠

PR #200, opened 2015-08-27
Added Event.isTrusted sample

beaufortfrancois:
Hello @jeffposnick,

Here's a simple sample to show how `Event.isTrusted` behaves. 

cc @RByers as well.


jeffposnick:
Thanks, @beaufortfrancois!

Could you refactor this example so that it doesn't use a third-party image? If you think it's important to keep using an image, you can include one of the icons from the `../images/` directory. Or you could just use a `<div>` with a background color, or whatnot.

It would also be nice to explain that the image is clickable, since I wouldn't immediately think to click on it unless I read through the code first.

event-istrusted/index.html
```diff
@@ -0,0 +1,52 @@
...
+    height: 50px;
+    border-radius: 50%;
+    border: 2px solid;
+    display: block;
+    margin: 0 auto;
+    cursor: pointer;
+  }
+</style>
+
+<p>The <code>isTrusted</code> read-only property of the <code>Event</code>
```
jeffposnick: `<a href="https://developer.mozilla.org/en-US/docs/Web/API/Event"><code>Event</code></a>`

event-istrusted/index.html
```diff
@@ -0,0 +1,52 @@
...
+interface is a <code>Boolean</code> that is <code>true</code> when the event was generated
+by a user action such as mouse click for instance, and <code>false</code> when
```
jeffposnick: Remove "for instance".

event-istrusted/index.html
```diff
@@ -0,0 +1,52 @@
...
+the event was created or modified by a script or dispatched via
+<code>dispatchEvent</code>.</p> <p>This new property is intended primarily for
```
jeffposnick: `<a href="https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent"><code>dispatchEvent</code></a>`

event-istrusted/index.html
```diff
@@ -0,0 +1,52 @@
...
+    display: block;
+    margin: 0 auto;
+    cursor: pointer;
+  }
+</style>
+
+<p>The <code>isTrusted</code> read-only property of the <code>Event</code>
+interface is a <code>Boolean</code> that is <code>true</code> when the event was generated
+by a user action such as mouse click for instance, and <code>false</code> when
+the event was created or modified by a script or dispatched via
```
jeffposnick: "the event was scripted or invoked via"

event-istrusted/index.html
```diff
@@ -0,0 +1,52 @@
...
+<code>dispatchEvent</code>.</p> <p>This new property is intended primarily for
+use by browser extensions, to determine if an event was dispatched by a script running
+in the main world or not.</p>
+
+<img id="keanu" src="Conspiracy-Keanu.jpg" title="Click me...">
+<button>BUTTON</button>
```
jeffposnick: Something like "Generate Trusted Event" would be a more descriptive title than "BUTTON".

beaufortfrancois:
@jeffposnick I've fixed everything so that it doesn't involve my dear Keanu anymore.
:crying_cat_face:

jeffposnick:
:+1: from me.

RByers:
Looks reasonable to me, thanks!   Thanks in particular for mentioning the extensions case - I'm sure some discussion of this will repeat the false believe that script in the main world can rely on `isTrusted` being un-tampered with by other script in the main world.

/cc @dtapuska

PR #202, opened 2015-08-31
Nicer HTTPS redirection for non-localhost

jeffposnick:
R: any/all

This matches the logic used at https://github.com/PolymerElements/platinum-https-redirect/blob/master/platinum-https-redirect.html, which is more in keeping with the actual restriction in place.

It explicitly detects `localhost` and its equivalents instead of assuming that anything not on port 80 is `localhost`.


addyosmani:
LGTM

PR #208, opened 2015-09-03
CSS Intrinsic Sizing sample

jeffposnick:
R: @addyosmani @gauntface @beaufortfrancois _et al._
CC: @cbiesinger

A sample showing off CSS Intrinsic Sizing keywords used on the width of a `<div>`.

@cbiesinger, it would be great to have you confirm that this demo sufficiently shows off the new functionality before we merge.

There's a deployed version at https://jeffy.info/samples/css-intrinsic-sizing/ that you can try out live.


gauntface:
LGTM from me.

PR #209, opened 2015-09-16
Update the prefetch SW example to use cache-busting parameters

jeffposnick:
R: @gauntface @wibblymat @PaulKinlan _et al._

The current example showing off a service worker that prefetches URLs and then uses those responses in a cache-first strategy runs the risk of locking users in to a stale response indefinitely. To mitigate this risk, this change appends a cache-busting URL parameter to the `fetch()` request made in the `install` handler. That way, we can be assured that the response to that initial request is fresh at the time the `install` handler is run.

(It's possible that the resource might change after the `install` handler is run, and coordinating those resource changes with the service worker cache versioning is a separate problem that's already alluded to in the comments.)

### DIFF
--- a/service-worker/prefetch/service-worker.js
+++ b/service-worker/prefetch/service-worker.js
@@ -34,9 +34,11 @@ var CURRENT_CACHES = {
 };
 
 self.addEventListener('install', function(event) {
+  var now = Date.now();
+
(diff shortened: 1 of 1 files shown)


wibblymat:
:+1:

PR #216, opened 2015-09-30
ES2015 Rest parameters sample

addyosmani:
r: @jeffposnick

### DIFF
--- a/rest-parameters-es6/index.html
+++ b/rest-parameters-es6/index.html
@@ -1,0 +1,81 @@
+---
+feature_name: Rest parameters (ES6)
+chrome_version: 47
+feature_id: 5937087912083456
+---
+
+<h3>Background</h3>
+<p>
+Rest parameters allow your functions to have variable number of arguments without using the arguments object. The rest parameter is an instance of Array so all the array methods just work.
+</p>
+
+{% include output_helper.html %}
+
+{% capture js %}
+
+// The Old ES5 way
+function listAnimals() {
+  var animals = Array.prototype.slice.call(arguments);
+  animals.forEach(function(animal) {
+    ChromeSamples.log(animal);
+  });
+}
+
+listAnimals('🐯', '🐰', '🐘', '🐴', '🐥');
+
+// New ES2015 way
+// This allows you to just define a rest parameter, a real array
+// which is a snapshot of the arguments passed to the function
+
+function listAnimals(...animals) {
+  animals.forEach(function(animal) {
+    ChromeSamples.log(animal);
+  });
+}
+
+// This approach becomes even simpler with ES6 fat arrows
+function listAnimalsArrows(...animals) {
+  animals.forEach((animal) => ChromeSamples.log(animal));
+}
+
+listAnimalsArrows('🐯', '🐰', '🐘', '🐴', '🐥');
+
+// Get the count of elements using the length property
+function fun1(...theArgs) {
+  ChromeSamples.log(theArgs.length);
+}
+
+fun1();  // 0
+fun1('🐩'); // 1
+fun1('🐩', '🐘', '🐥'); // 3
+
+// Collect arguments from the second one to the end
+
+function multiply(multiplier, ...theArgs) {
+  return theArgs.map(function (element) {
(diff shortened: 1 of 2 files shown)


rest-parameters-es6/index.html
```diff
@@ -0,0 +1,83 @@
+---
+feature_name: Rest parameters (ES2015)
+chrome_version: 47
+feature_id: 5937087912083456
+---
+
+<h3>Background</h3>
+<p>
+Rest parameters allow your functions to have variable number of arguments without using the arguments object. The rest parameter is an instance of Array so all the array methods just work.
```
jeffposnick: Rest parameters → [Rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)

arguments → [`arguments`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments)

rest-parameters-es6/index.html
```diff
@@ -0,0 +1,83 @@
...
+</p>
```
jeffposnick, line 10: Can you mention in the Background section that there can only be a single rest parameter for a given function? It's obvious when you know about the feature, but I could imagine some folks newer folks wondering what happens if you try `function foo(...first, ...second)`

rest-parameters-es6/index.html
```diff
@@ -0,0 +1,83 @@
...
+  animals.forEach(function(animal) {
+    ChromeSamples.log(animal);
+  });
+}
+
+listAnimalsES2015('🐯', '🐰', '🐘', '🐴', '🐥');
+
+// This approach becomes even simpler with ES2015 fat arrows
+function listAnimalsArrows(...animals) {
+  animals.forEach((animal) => ChromeSamples.log(animal));
```
jeffposnick: (animal) → animal

rest-parameters-es6/index.html
```diff
@@ -0,0 +1,83 @@
...
+}
+
+listAnimalsArrows('🐯', '🐰', '🐘', '🐴', '🐥');
+
+// Get the count of elements using the length property
+function fun1(...theArgs) {
```
jeffposnick: fun1 → countArguments (or something else that's more descriptive)

jeffposnick:
Reviewed. Thanks!

rest-parameters-es6/index.html
```diff
@@ -0,0 +1,83 @@
...
+}
+
+ChromeSamples.log(sortRestArgs(5,3,7,1)); // shows 1,3,5,7
+
+function sortArguments() {
+  var sortedArgs = arguments.sort();
+  return sortedArgs; // this will never happen
+}
+
+// throws a TypeError: arguments.sort is not a function
```
beaufortfrancois, line 82: To be fair here, we could mention `Array.from`:

``` js
function sortArguments() { 
  return Array.from(arguments).sort(); 
};
```

addyosmani:
Thanks for reviewing. Attempted to address comments. PTAL.

rest-parameters-es6/index.html
```diff
@@ -0,0 +1,93 @@
...
+chrome_version: 47
+feature_id: 5937087912083456
+---
+
+<h3>Background</h3>
+<p>
+<a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters">Rest parameters</a> allow your functions to have variable number of arguments without using the <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments">arguments</a> object. The rest parameter is an instance of Array so all the array methods just work.
+</p>
+
+<p>Note that there can only be a single rest parameter for a given function. This means that `listAnimals(...cats, ...dogs)` will not work, whilst `listAnimals(...cats)` will work just fine.</p>
```
jeffposnick: The backtick escapes need to be `<code>` here, since it's just plain HTML.

jeffposnick:
LGTM after the change to `<code>` is made.

addyosmani:
Correction made. Thanks for reviewing!

PR #217, opened 2015-10-06
Redirect to https://www.chromestatus.com/samples

jeffposnick:
R: @ebidel 

Change the https://googlechrome.github.io/samples/ top-level redirect to point to the new https://www.chromestatus.com/samples page.


ebidel:
LGTM

PR #219, opened 2015-10-06
Adds a 'error' event handler to window

### LINKED ISSUE #218: Array.prototype.includes sample does not trap errors so output is empty
https://googlechrome.github.io/samples/array-includes-es7/index.html has the "Live Output" box empty in a browser not supporting `Array.prototype.includes`, because the first example (`if (basketOfApples.includes('🌰'))`) fails (because of an `Uncaught TypeError: basketOfApples.includes is not a function`) and so the remainder of the code is not executed. Perhaps it would be useful if errors were trapped and some sort of "failure" message added to the `Live Output` section, so that one can see that one's current browser does not support this sample (and one can tell the difference between "my browser tried to do this and failed because it's not new enough" and "the code is broken or just doesn't work").
(This is likely not specific to this particular sample and is probably a more general class of error in the samples as a whole.)

jeffposnick:

This isn't a fool-proof solution, but for any samples that include the output helper template, this will end up logging something meaningful when you try out a feature that's not supported in your current browser.

image

Closes #218

### DIFF
--- a/rest-parameters-es6/index.html
+++ b/rest-parameters-es6/index.html
@@ -80,7 +80,11 @@ function sortArguments() {
 }
 
 // throws a TypeError: arguments.sort is not a function
-ChromeSamples.log(sortArguments(5,3,7,1));
+try {
+  ChromeSamples.log(sortArguments(5,3,7,1));
+} catch(error) {
+  console.warn(error);
+}
 
 /*
 We could alternatively have used Array.from here:
--- a/rest-parameters-es6/index.html
+++ b/rest-parameters-es6/index.html
@@ -83,7 +83,7 @@ function sortArguments() {
 try {
   ChromeSamples.log(sortArguments(5,3,7,1));
 } catch(error) {
-  console.warn(error);
+  ChromeSamples.log(error.message);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
I like the general idea but we may want to rephrase wording as the https://googlechrome.github.io/samples/rest-parameters-es6/index.html sample with that specific patch shows `Uncaught TypeError: arguments.sort is not a function (Your browser does not currently support this feature.)`.

addyosmani:
Like the direction we're taking with this. Agree with Francois that improving the wording would be desirable before landing this. Thanks for the PR, @jeffposnick!

jeffposnick:
I can't think of any other instance in which we intentionally trigger an exception in a sample that is otherwise supported in a given browser.

Here's a slight update to the wording, but more to the point, an explicit `try` / `catch` around that intentional exception that just logs it to the `console`.

rest-parameters-es6/index.html
```diff
@@ -80,7 +80,11 @@
 }
 
 // throws a TypeError: arguments.sort is not a function
-ChromeSamples.log(sortArguments(5,3,7,1));
+try {
+  ChromeSamples.log(sortArguments(5,3,7,1));
+} catch(error) {
+  console.warn(error);
```
beaufortfrancois: We may want to add this error in `ChromeSamples.log`.

beaufortfrancois:
LGTM except nit.

PR #220, opened 2015-10-06
put new.target sample in strict mode

brendankenny:
to work in current Chrome while waiting for sloppy-mode block scoping. [The sample](https://googlechrome.github.io/samples/new-target-es6/index.html) currently doesn't work without `#enable-javascript-harmony` enabled.


jeffposnick:
Thanks, @brendankenny!

PR #222, opened 2015-10-07
strict mode

hemanth:
Or else we get an error `Uncaught SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode`


hemanth:
Ah, noticed #220 closing this.

PR #224, opened 2015-10-08
Input Device Capabilities example

jeffposnick:
R: @rbyers
CC: @addyosmani @beaufortfrancois @gauntface 

This is a sample for the [Input Device Capabilities functionality](https://www.chromestatus.com/feature/5681847971348480) found in M47.

@rbyers, it would be great if you could confirm that it's properly demonstrating the use case and motivation. As far as I can tell, the only property available in M47 is `firesTouchEvents`.

You can view a live version of the sample at https://jeffy.info/samples/input-device-capabilities/


input-device-capabilities/index.html
```diff
@@ -0,0 +1,36 @@
...
+<h3>Background</h3>
+<p>
+  The Input Device Capabilities API provides information about the capabilities of the physical
+  device responsible for generating an event.
+</p>
+<p>
+  This sample illustrates using <code>UIEvent.sourceCapabilities</code> (via a <code>click</code>
+  <code>UIEvent</code>) to obtain the capabilities of the device that triggered the click, and
+  then checks the <code>firesTouchEvents</code> property to determine whether that device can also
+  create a <a href="http://www.w3.org/TR/touch-events/#touchevent-interface"><code>TouchEvent</code></a>.
```
beaufortfrancois: We may want to add a comment like _This doesn't necessarily mean the device is a touch screen. For example, stylus and mouse devices typically generate touch events on mobile browsers._

beaufortfrancois:
LGTM expect nit.

input-device-capabilities/index.html
addyosmani: Good call, @beaufortfrancois!

addyosmani:
LGTM2 minus one nit about linking out `Input Device Capabilities API`

jeffposnick:
Thanks; feedback addressed. It would still be great to hear from @rbyers to confirm that this is demonstrating everything that's making it in to M47.

RByers:
Sorry for the delay (was travelling). Yep, it's very simple ATM.  LGTM.

Note that @paulkinlan is working on a post about this which will be a better reference than the old html5rocks article you site.

PaulKinlan:
Yep, its currently in a branch on WF waiting to deploy.

RByers:
BTW: simple real-world use case is a button that styles itself depressed on 'touchstart' and real 'mousedown' but does nothing on the "fake" 'mousedown' you get when tapping on a touchscreen.

jeffposnick:
Thanks for that extra context, @RByers. I've tried to address it.

I've also moved the sample's JS to `demo.js`, to match the new best practice.

The latest code is redeployed at https://jeffy.info/samples/input-device-capabilities/

RByers:
LGTM but note that your introduction still references "click" when you're
in fact looking at other events now.

jeffposnick:
Good catch. Thanks!

PR #227, opened 2015-10-13
Base template/style tweaks

jeffposnick:
R: @ebidel @gauntface @beaufortfrancois @addyosmani

This adds a link to the new https://www.chromestatus.com/samples landing page at the top of each templated sample's page.

It also tweaks the shared styles to hide the output helper elements when they're empty, which fixes a small visual issue with the margins.


gauntface:
Will the display none result in a pop when the log has something in or is it not too harsh?

The chrome status link looks all ok.

jeffposnick:
There's already popping/resizing going on as each additional log entry expands the size of the output area. This PR should make the initial, pre-logging/status margins look cleaner.

Compare
https://jeffy.info/samples/report-validity/index.html (with PR) to
https://googlechrome.github.io/samples/report-validity/index.html (without PR)

gauntface:
I can't tell the different on mobile so I'm going to say lgtm

addyosmani:
Changes LGTM2.

PR #228, opened 2015-10-13
Templatization of older samples

jeffposnick:
R: @gauntface, all

I'm gradually moving older samples over to the new templates, especially ones with inlined JavaScript. This helps with consistency and means we'll be able to lint the standalone JS.

Expect further PRs to trickle in as I'll do them in batches.


gauntface:
LGTM

PR #229, opened 2015-10-13
NotificationOptions.requireInteraction

wibblymat:
Also moved the existing vibrate option sample with it. Not sure if this is acceptable given that it will break any incoming links, but I don't think that sample was ever linked to in the first place.

R: @jeffposnick @gauntface


jeffposnick:
Thanks, @wibblymat! Works as expected, and I agree that breaking out the existing sample (which I can't find any direct links to) from the `vibration` directory makes sense.

One thing that's missing is a top-level `index.html` like https://github.com/GoogleChrome/samples/blob/gh-pages/web-bluetooth/index.html

I'm just going to push a commit that adds it, and then merge in the interest of getting this published.

PR #231, opened 2015-10-14
Improve #log text in small viewports

beaufortfrancois:
R=@jeffposnick

On mobile devices, I can't read the error messages in #log.
This small fix addresses this.


jeffposnick:
:+1:

PR #235, opened 2015-10-26
Added web-bluetooth characteristic properties sample

beaufortfrancois:
This sample simply features new Bluetooth characteristics now available in Canary :bird: 

R=@jeffposnick


web-bluetooth/characteristic-properties.js
```diff
@@ -0,0 +1,32 @@
...
+  if (serviceUuid.startsWith('0x')) {
+    serviceUuid = parseInt(serviceUuid, 16);
+  }
+
+  var characteristicUuid = document.getElementById('characteristic').value;
+  if (characteristicUuid.startsWith('0x')) {
+    characteristicUuid = parseInt(characteristicUuid, 16);
+  }
+
+  log('Requesting Bluetooth Device...');
```
jeffposnick, line 12: Shouldn't this be `ChromeSamples.log()`?

beaufortfrancois, line 12: I have `log = ChromeSamples.log;` to make it look nicer and shorter ;)

jeffposnick:
:+1: in general.

Based on your understanding of the API, would it make sense to break out the `<option value="alert_notification">`, etc. values into a separate partial template that you'd then include in multiple samples? I'm starting to see repetition across multiple examples. My lack of familiarity with the API makes me unsure of how boilerplate those values are, though.

beaufortfrancois:
I've added files for each datalist as you suggested.
Thank you!

jeffposnick:
:+1:

PR #238, opened 2015-10-27
Switched to https://www.chromestatus.com instead of http

beaufortfrancois:
In order to avoid an unnecessary HTTP 302 Redirect, I ran `git grep -l 'http://www.chromestatus.com' | xargs sed -i 's/http:\/\/www.chromestatus.com/https:\/\/www.chromestatus.com/g'` in the repository.

R=@jeffposnick


jeffposnick:
Thanks!

PR #239, opened 2015-10-27
Switched to https://developers.google.com/web/updates/ when possible

beaufortfrancois:
Here's the command I ran there:
`git grep -l 'http://updates.html5rocks.com' | xargs sed -i 's/http:\/\/updates.html5rocks.com/https:\/\/developers.google.com\/web\/updates/g'`

R=@jeffposnick


PaulKinlan:
I'll let Jeff review, but this gets a +1 from me.

jeffposnick:
:+1: Thanks!

PR #240, opened 2015-10-27
Removed service existence check

beaufortfrancois:
Since https://codereview.chromium.org/1378013002, we don't need to check for service existence anymore.

R=@jeffposnick


jeffposnick:
:+1: I will take your word on that.

PR #241, opened 2015-10-28
Added ability to override samples URL for a sample

beaufortfrancois:
I'd love if we could customize the "browse samples" link at the very top.
It makes more sense to me to visit "Web bluetooth samples" than chromestatus.com/samples when I'm on a Web Bluetooth sample.

R=@jeffposnick


jeffposnick:
I'm not in favor of overriding the default template's "Browse Samples" link. @ebidel did a great job with https://www.chromestatus.com/samples, and I'd like for developers who stumble across any of our individual samples to see a link pointing back to it. I don't see why we should special-case developers who come across our Web Bluetooth samples.

If you're concerned about developers stumbling across an individual Web Bluetooth sample page and not realizing that there are other Web Bluetooth samples, then I'd suggest just including a paragraph at the top of each sample with some text and a link pointing to `/web-bluetooth/index.html`.

beaufortfrancois:
I will include a paragraph then for each of the web bluetooth samples.

jeffposnick:
Cool. I think that will work best. :+1:

PR #242, opened 2015-10-29
Added web-bluetooth samples intro

beaufortfrancois:
After discussing about it at https://github.com/GoogleChrome/samples/pull/241, here it is!

@jeffposnick


jeffposnick:
:+1:

PR #243, opened 2015-11-03
Replaced deprecated instanceId

beaufortfrancois:
R=@jeffposnick

See https://codereview.chromium.org/1409903004/


jeffposnick:
:+1:

PR #244, opened 2015-11-03
Added bluetooth characteristic notifications sample

beaufortfrancois:
R=@jeffposnick
screenshot 2015-11-03 at 9 49 59 am

### DIFF
--- a/web-bluetooth/notifications.html
+++ b/web-bluetooth/notifications.html
@@ -1,0 +1,55 @@
+---
+feature_name: Web Bluetooth / Notifications
+chrome_version: 48
+feature_id: 5264933985976320
+---
+
+{% include_relative _includes/intro.html %}
+
+<p>This sample illustrates the use of the Web Bluetooth API to start and stop
+characteristic notifications from a nearby Bluetooth Low Energy Device. You may
+want to try this demo with the BLE Peripheral Simulator App
+from the <a target="_blank"
+href="https://play.google.com/store/apps/details?id=io.github.webbluetoothcg.bletestperipheral">Google
+Play Store</a>.</p>
+
+<form>
+  <input id="service" type="text" list="services" autofocus placeholder="Bluetooth Service">
+  <input id="characteristic" type="text" list="characteristics" placeholder="Bluetooth Characteristic">
+  <button id="startNotifications">Start notifications</button>
+  <button id="stopNotifications">Stop notifications</button>
+</form>
+
+{% include_relative _includes/datalist-services.html %}
+{% include_relative _includes/datalist-characteristics.html %}
+
+{% include output_helper.html %}
+
+{% include js_snippet.html filename='notifications.js' %}
+
+<script>
+  document.querySelector('#startNotifications').addEventListener('click', function(event) {
+    event.stopPropagation();
+    event.preventDefault();
+
+    if (!navigator.bluetooth) {
+      ChromeSamples.setStatus('Web Bluetooth API is not available.\n' +
+        'Please make sure the Web Bluetooth flag is enabled.');
+    } else {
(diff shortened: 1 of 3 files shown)


web-bluetooth/notifications.html
```diff
@@ -0,0 +1,55 @@
...
+{% include_relative _includes/datalist-characteristics.html %}
+
+{% include output_helper.html %}
+
+{% include js_snippet.html filename='notifications.js' %}
+
+<script>
+  document.querySelector('#startNotifications').addEventListener('click', function(event) {
+    event.stopPropagation();
+    event.preventDefault();
```
jeffposnick, line 33: I don't think there's a benefit of calling `stopPropagation()`/`preventDefault()` here and elsewhere, as the DOM element is just a normal `<button>`.

web-bluetooth/notifications.js
```diff
@@ -0,0 +1,47 @@
...
+      log('> Notifications stopped');
+      myCharacteristic.removeEventListener('characteristicvaluechanged', handleNotifications);
+    });
+  }
+}
+
+function handleNotifications(event) {
+  var buffer = event.target.value;
+  var data = new DataView(buffer);
+  for (var i = 0, a = []; i < data.byteLength; i++) {
```
jeffposnick, line 47: Could you include a comment about what the source data will look like? A series of bytes, I know, but maybe there's some additional context? And then a comment about how you're string-ifying it here, but how for a "real" use case you might just process the raw bytes?

Right now, the loop lacks context.

jeffposnick:
Reviewed. Thanks, @beaufortfrancois!

web-bluetooth/notifications.js
```diff
@@ -0,0 +1,47 @@
+var myCharacteristic;
```
g-ortuno: any reason why you are not using 'use strict'?

web-bluetooth/notifications.html
beaufortfrancois, line 33: Since `<button>` is inside a `<form>` so that user can simply press Enter to validate it, I need to call `stopPropagation()/preventDefault()`.

web-bluetooth/notifications.js
beaufortfrancois, line 47: Done.

web-bluetooth/notifications.js
beaufortfrancois: Not a good one I can find out ;)
I've fixed this at 360f5a991736426e8c2b18fbab772729a2137558

beaufortfrancois:
@jeffposnick @g-ortuno Feedback addressed!
Thanks for the review.

web-bluetooth/notifications.html
jeffposnick, line 33: Oops, missed that detail. Thanks for clarifying!

jeffposnick:
Hmm, the Travis CI build is now unhappy about `log = ChromeSamples.log;` being defined in `notifications.html` and not in `notifications.js`. Would you mind moving it to the top of `notifications.js` to turn things green again?

beaufortfrancois:
Since this has nothing to do with the web-bluetooth samples, I'm reluctant to add `log = ChromeSamples.log;` in `notifications.js`.
This has never been a problem until now and I wonder if we could make an exception for it.

jeffposnick:
Yeah, I'm not sure why this specific instance is getting flagged by our `jshint` task as opposed to the same logic in the other files. If you want to add in explicit comments to disable that `jshint` rule for the file, that's okay, but it may end up being the same level of extra noise as it would if you just explicitly moved the `log = ChromeSamples.log;` definition into the file.

jeffposnick:
Okay, #246 takes care of it.

PR #245, opened 2015-11-04
Added missing 'use strict' in bluetooth samples

beaufortfrancois:
R= @jeffposnick 

Based on @g-ortuno feedback, you'll find missing `"use strict"` for web-bluetooth samples.


jeffposnick:
:+1:

PR #246, opened 2015-11-04
Added custom log function to .jshintrc

beaufortfrancois:
R=@jeffposnick

Would that work for you?
Background: https://github.com/GoogleChrome/samples/pull/244


jeffposnick:
:+1:

PR #250, opened 2015-11-13
Fix off by 1 error in generators example

anka-213:
The 300-px example printed "undefined" between each full cycle of "Lorem ipsum".


jeffposnick:
Thanks! Can you please read through and, assuming you agree, sign the Google CLA?

anka-213:
Done.

jeffposnick:
Thanks!

PR #254, opened 2015-11-20
Custom Offline Page sample revamp

jeffposnick:
R: @gauntface @addyosmani @beaufortfrancois 

A number of the service worker samples are up for a revamp.

First up is this PR, the highlights of which include:
- Moving to the templating setup, and adding the `service-worker.js` code inline in the rendered HTML, but not executing it, via the new `{% include ... displayonly=true %}` option.
- Switching to :arrow_right: functions.
- Conditional cache-busting, which is skipped on browsers that support `Request.cache` set to `reload`.
- Explicit reference to the need for cache versioning in the comments.

You can view a live version of this change at https://jeffy.info/samples/service-worker/custom-offline-page/


addyosmani:
Minor comment but otherwise LGTM. If you end up keeping as is, I'd suggest we squash commits for a clean land of this work.

jeffposnick:
Hmm, I inadvertently used my fork's `gh-pages` branch for this PR instead of the dedicated branch for the new functionality, so it's picking up some extra commits.

I'd like to keep in the future-facing piece of code, so it sounds like I have the LGTM. Rather than squashing commits here, I'm going to refile the PR against my branch and then merge.

jeffposnick:
...except I deleted my fork's branch last week. So re-opening this, and I'll merge post-squash.

PR #255, opened 2015-11-24
KeyboardEvent.code attribute sample

jeffposnick:

Sample code for https://www.chromestatus.com/feature/5228092293382144

Live version available for preview at https://jeffy.info/samples/keyboardevent-code-attribute/

### DIFF
--- a/keyboardevent-code-attribute/README.md
+++ b/keyboardevent-code-attribute/README.md
@@ -1,0 +1,5 @@
+KeyboardEvent code Attribute Sample
+===
+See https://googlechrome.github.io/samples/keyboardevent-code-attribute/index.html for a live demo.
+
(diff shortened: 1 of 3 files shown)


addyosmani:
PR LGTM.

beaufortfrancois:
I wonder if we should call `event.preventDefault();` as well so that user can fully interact with the sample without to worry about typing weird keys combos.

jeffposnick:
It didn't seem necessary when I was playing around with the demo.

Ultimately this new attribute is focused on individual keypresses rather than modifier keys. If someone wants to type Control/Apple+W, I'd rather it closed the tab like it would by default rather than hijacking things.

beaufortfrancois:
OK.

PR #256, opened 2015-11-24
FontFaceSet sample

jeffposnick:
CC: @toyoshim, who I believe worked on the Chrome implementation.

This is a sample for the `FontFaceSet` interface in M48, https://www.chromestatus.com/feature/4594172182921216

I'm not 100% sure that it illustrates all the best practices properly, so it would be great if @toyoshim could offer some advice/review. It's also attempting to show off a few different things at once (loading custom `FontFace`s via both JS and CSS, hiding text until the fonts have loaded, iterating over all the loaded `FontFace`s) so let me know if you feel like I should cut something out to make it more focused.

You can view a live version at https://jeffy.info/samples/font-face-set/

### DIFF
--- a/font-face-set/README.md
+++ b/font-face-set/README.md
@@ -1,0 +1,5 @@
+CSS Font Loading API's FontFaceSet Sample
+===
+See https://googlechrome.github.io/samples/font-face-set/index.html for a live demo.
+
+Learn more at https://www.chromestatus.com/feature/4594172182921216
--- a/font-face-set/demo.js
+++ b/font-face-set/demo.js
@@ -1,0 +1,26 @@
+// Load Bitter via JavaScript, as an alternative to CSS's @font-face rule.
+var bitterFontFace = new FontFace('Bitter', 'url(https://fonts.gstatic.com/s/bitter/v7/HEpP8tJXlWaYHimsnXgfCOvvDin1pK8aKteLpeZ5c0A.woff2)');
+document.fonts.add(bitterFontFace);
+
+// The .ready promise resolves when all fonts that have been previously requested
+// are loaded and layout operations are complete.
+document.fonts.ready.then(function() {
+  // Our CSS sets visibility: hidden on the oxygen class initially,
+  // so let's make it visible now.
+  // This technique avoids displaying text until the font it requires is loaded.
+  // An alternative approach is used with the text styled with the bitter class,
+  // which is visible from the start using the default font,
+  // and then switches to the custom Bitter font once that's loaded.
+  document.querySelector('.oxygen').style.visibility = 'visible';
+
+  ChromeSamples.log('There are', document.fonts.size, 'FontFaces loaded.\n');
+
+  // document.fonts has a Set-like interface. Here, we're iterating over its values.
+  for (var fontFace of document.fonts.values()) {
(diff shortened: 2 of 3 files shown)


beaufortfrancois:
Maybe adding a log font loading progress would help to visualize the process.
Otherwise LGTM.

jeffposnick:
That's an interesting idea. I don't know very much about this API, so I'm poking around a bit to see what's available...

There are `loading`/`loadingdone`/`loadingerror` events exposed on the `FontFaceSet`, but when those fire, I don't see anything in the corresponding `FontFaceSetLoadEvent` parameter indicating which underlying `FontFace` the event applies to. So I'm not sure it offers more of a benefit vs. just using the `document.fonts.ready` promise to wait for the whole set of fonts to load.

There aren't any events exposed on the individual `FontFace` objects, though there is a `.loaded` promise you can wait for. The problem with that is I'm not sure how to construct a new `FontFace` object corresponding to the font loaded via `@font-face` in CSS, so I would only be able to show waiting on `.loaded` for the font loaded via JavaScript.

Maybe @toyoshim has some ideas?

addyosmani:
Sample LGTM. I wasn't able to find additional benefits to using more than `document.fonts.ready` to wait on the entire set of fonts to load vs. using the additional events exposed, but this might just be my poor reading of the material around this feature.

gauntface:
I'm a bit confused at what this demo is trying achieve.

Oxygen is hidden but the font is defined in CSS. Bitter is visible but loaded via JS.

Why toggle oxygen visibility?

Why Load one in CSS the other in JS?

How does blocking in CSS affect visibility of fonts with all of this?

Is it possible to get a callback when a specific font is loaded?

This still seems to exhibit the same behaviour as the web has today (i.e. it will just display when all fonts are loaded.

Is it worth:
- removing any unneccessary steps (i.e. toggling visibility)
- Splitting out into multiple small demos (i.e. here the font is loaded via CSS and we know when it's upgraded by callback X. Here the font is loaded via JS and we get a callback when its loaded).

jeffposnick:
Yeah, the sample does try to cover a lot.

I toggle the visibility inside the `.ready` promise because that seems to be the intended use of that promise. It's the example in the [spec](https://drafts.csswg.org/css-font-loading/#font-load-event-examples) as well. That does seem legitimate to me, but I could explain it better in the sample—it prevents you from seeing the text temporarily in the default font and then having it flash into the custom font.

I can remove loading the font via CSS, as it's not part of the new functionality. I thought it would be nice to have multiple fonts loaded, since one of the other pieces of new functionality is the ability to iterate over all the custom fonts on the page.

It's possible to use a promise to check one a single font is loaded via JavaScript, but like I mentioned in my previous comment, I don't see that as being possible with the font loaded via CSS.

I could change the sample to load multiple fonts via JavaScript and log once each is loaded, and then log again once the `.ready` promise is resolved. Would that address the feedback?

gauntface:
Totally would yes :)

jeffposnick:
PTAL.

gauntface:
LGTM

toyoshim:
Ouch. I rarely checked github notifications.
Please ping me by mail or something if you ask me here and can not get a resoponse in the future.

PR #257, opened 2015-11-24
Network Information API updates

jeffposnick:
CC: @jkarlin

This updates our existing Network Information API sample from the M38 era to include the new M48 functionality (https://www.chromestatus.com/feature/6338383617982464).

A live version of the updates can be found at https://jeffy.info/samples/network-information/, but as with the initial sample, you'll need an Android/iOS/Chrome OS device for the output to be meaningful.

### DIFF
--- a/network-information/index.html
+++ b/network-information/index.html
@@ -1,95 +1,33 @@
-<!doctype html>
-<!--
-Copyright 2014 Google Inc. All Rights Reserved.
+---
+feature_name: Network Information API
+chrome_version: 48
+feature_id: 6338383617982464
+---
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+<h3>Background</h3>
+<p>
+  This demo illustrates the use of the
+  <a href="http://w3c.github.io/netinfo/">Network Information API</a>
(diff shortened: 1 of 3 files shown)


beaufortfrancois:
LGTM :+1:

network-information/index.html
```diff
@@ -1,95 +1,33 @@
...
+<p>
+  This demo illustrates the use of the
+  <a href="http://w3c.github.io/netinfo/">Network Information API</a>
+  to retrieve the current network type and monitor for network type changes.
+</p>
+
+<p>
+  The Network Information API is
+  <a href="https://code.google.com/p/chromium/issues/detail?id=368358#c18">not currently available</a>
+  on all platforms. For best results, try this demo using Chrome on Android, iOS, or ChromeOS.
```
jkarlin: The iOS bit isn't true. We don't support net info on ios unfortunately. Otherwise LGTM.

jeffposnick: Thanks for pointing that out!

gauntface:
Breaks on browsers that don't support it.

jeffposnick:
See my comment at https://github.com/GoogleChrome/samples/pull/258#issuecomment-159922809 

Is that not sufficient?

PR #258, opened 2015-11-24
IDB getAll() sample

jeffposnick:
CC: @inexorabletash

This implements a basic sample to go along with the new M48 IndexedDB "batch get" methods (https://www.chromestatus.com/feature/6537756637396992).

I'm using a light Promise wrapper around getting the IndexedDB instance, but I don't think that's too distracting.

You can view a deployed version of this sample at https://jeffy.info/samples/idb-getall/

### DIFF
--- a/idb-getall/README.md
+++ b/idb-getall/README.md
@@ -1,0 +1,5 @@
+IndexedDB getAll() Methods Sample
+===
+See https://googlechrome.github.io/samples/idb-getall/index.html for a live demo.
+
+Learn more at https://www.chromestatus.com/feature/6537756637396992
--- a/idb-getall/demo.js
+++ b/idb-getall/demo.js
@@ -29,6 +29,8 @@ document.querySelector('#add').addEventListener('click', function(event) {
     var transaction = db.transaction(STORE_NAME, 'readwrite');
     var objectStore = transaction.objectStore(STORE_NAME);
     objectStore.put(Date.now());
+  }).catch(function(error) {
+    ChromeSamples.setStatus(error);
   });
 });
 
@@ -42,5 +44,7 @@ document.querySelector('#display').addEventListener('click', function(event) {
       ChromeSamples.setStatus('There are ' + timestamps.length + ' timestamps saved in IndexedDB: '
         + timestamps.join(', '));
     };
+  }).catch(function(error) {
+    ChromeSamples.setStatus(error);
   });
 });
--- a/idb-getall/demo.js
+++ b/idb-getall/demo.js
@@ -1,50 +1,41 @@
+var addButton = document.querySelector('#add');
+var displayButton = document.querySelector('#display');
+
 var DB_NAME = 'timestamps';
 var DB_VERSION = 1;
 var STORE_NAME = 'store';
-var dbInstance;
+var db;
 
-// Simple Promise-based wrapper around getting an IndexedDB instance.
-function getDB() {
-  if (dbInstance) {
-    return Promise.resolve(dbInstance);
-  }
-
-  return new Promise(function(resolve) {
-    var request = indexedDB.open(DB_NAME, DB_VERSION);
-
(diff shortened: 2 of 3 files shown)


beaufortfrancois:
To be honest, I was confused about the Promise wrapper. I think something simpler as below would help developers to focus more on the `getAll` method.

``` diff
-var DB_NAME = 'timestamps';
-var DB_VERSION = 1;
 var STORE_NAME = 'store';
-var dbInstance;
-
-// Simple Promise-based wrapper around getting an IndexedDB instance.
-function getDB() {
-  if (dbInstance) {
-    return Promise.resolve(dbInstance);
-  }
-
-  return new Promise(function(resolve) {
-    var request = indexedDB.open(DB_NAME, DB_VERSION);
-
-    request.onupgradeneeded = function() {
-      request.result.createObjectStore(STORE_NAME, {autoIncrement: true});
-    };
-
-    request.onsuccess = function() {
-      dbInstance = request.result;
-      resolve(dbInstance);
-    };
-  });
-}
+var db;

+var request = indexedDB.open('timestamps', 1);
+request.onsuccess = function() {
+  db = request.result;
+};
+request.onupgradeneeded = function() {
+  db.createObjectStore(STORE_NAME, { autoIncrement: true});
+};

 document.querySelector('#add').addEventListener('click', function(event) {
-  getDB().then(function(db) {
   var transaction = db.transaction(STORE_NAME, 'readwrite');
   var objectStore = transaction.objectStore(STORE_NAME);
   objectStore.put(Date.now());
-  }).catch(function(error) {
-    ChromeSamples.setStatus(error);
-  });
 });

 document.querySelector('#display').addEventListener('click', function(event) {
-  getDB().then(function(db) {
   var transaction = db.transaction(STORE_NAME, 'readonly');
   var objectStore = transaction.objectStore(STORE_NAME);
   // IDBObjectStore.getAll() will return the full set of items in our store.
   objectStore.getAll().onsuccess = function(event) {
     var timestamps = event.target.result;
-      ChromeSamples.setStatus('There are ' + timestamps.length + ' timestamps saved in IndexedDB: '
-        + timestamps.join(', '));
+    ChromeSamples.setStatus('There are ' + timestamps.length +
+        ' timestamps saved in IndexedDB: ' + timestamps.join(', '));
   };
-  }).catch(function(error) {
-    ChromeSamples.setStatus(error);
-  });
 });
```

jeffposnick:
It looks like there's a race condition in your proposed modifications, in that the `click` event listeners might be triggered before `db` is assigned to a fully functional IndexedDB instance.

The promise wrapper could be overkill, sure. What I could do is start out with the buttons initially disabled, and only enable them inside of `request.onupgradeneeded`, which should be safe enough for a sample.

beaufortfrancois:
Yeah you're right. 
Enabling buttons should be enough then.

jeffposnick:
PTAL.

beaufortfrancois:
LGTM except nit.

gauntface:
Worth adding a check when the sample is run on a browser that doesn't support this API?

jeffposnick:
@gauntface, there's a global [`error` handler](https://github.com/GoogleChrome/samples/blob/gh-pages/_layouts/default.html#L42) in place to avoid having to add in feature checks in all the samples. Here's what it should look like on unsupported browsers:

image

(The earlier version of this samples didn't trigger the handler, since a failure involved a promise rejection rather than an exception, but the modified version without promises will trigger it.)

gauntface:
My main thinking is that there may be developers that want to feature detect.

jeffposnick:
Given that there's a clear fallback behavior to use, sure, I'll update the sample to use that if `getAll()` isn't supported.

jeffposnick:
@gauntface PTAL. I've modified the sample to check for `getAll()` support, and fallback to cursors if it's not supported.

gauntface:
LGTM

PR #259, opened 2015-11-25
Service worker fallback-response sample modernization

jeffposnick:

Templatizes and simplifies the fallback response SW sample.

I'm using `fetch()` from the main page, which wasn't supported until Chrome 42, so I bumped up the required version info accordingly.

The sample now shows off doing conditional `FetchEvent.respondWith()` and constructing a `Response` object using a string. It also uses `skipWaiting()` and `clients.claim()`, so that we don't have to tell users to refresh their browser page the first time they visit the demo.

The updated code is deployed to https://jeffy.info/samples/service-worker/fallback-response/


addyosmani:
:+1: from me. Will defer to Matt or Francois for further comments.

PR #263, opened 2015-11-30
Correct e.g. ("for example") to i.e. ("in other words")

jpmedley:


jeffposnick:
That's intentionally "e.g."

There are multiple browsers each with their own flavor of developer tools, and I was just using the Chrome DevTools as one example.

It is missing a trailing '.', though, so if you want to change `e.g` to `e.g.` via this PR, I'm in favor of that.

jpmedley:
I know this isn't exactly what you asked. In my experience people frequently get these mixed up. That's why I misread it. That and the fact that the context doesn't clearly support one interpretation over the other. Using English clarifies it.

jeffposnick:
:+1:

PR #264, opened 2015-12-03
Service worker "kill switch" sample

jeffposnick:
R: @addyosmani @gauntface @wibblymat @jakearchibald @slightlyoff

From time to time folks ask about implementing a service worker "kill switch". I've tried to distill some best practices into a modular sample, and include information about why adding in a "kill switch" might not be necessary (shift-reloads, 24 hour max SW.js cache lifetime).

Curious to hear whether folks find this sample useful and sufficiently motivated/explained.

You can play around with a live version of the code at https://jeffy.info/samples/service-worker/kill-switch/

### DIFF
--- a/service-worker/kill-switch/index.html
+++ b/service-worker/kill-switch/index.html
@@ -1,0 +1,61 @@
+---
+feature_name: "Service Worker Sample: Kill Switch"
+chrome_version: 45
+feature_id: 6561526227927040
+---
+
+<h3>Background</h3>
+<p>
+  This samples demonstrates one approach to implementing a service worker
+  "<a href="https://en.wikipedia.org/wiki/Kill_switch">kill switch</a>". This pages registers a
+  service worker that has an undesirable <code>fetch</code> handler, which unconditionally
+  returns useless HTML in response to every fetch request. This simulates the "worst case scenario"
+  for a buggy or poorly implemented service worker. <strong>Once you reload this page, you'll see the
+  undesirable service worker in action.</strong>
+</p>
+
+<p>
+  Before getting into the technical details of the kill switch, it's worth pointing out that even
+  in this "worst case scenario", browsers offer built-in remediations that might prove sufficient.
+  One remediation is to
+  <a href="https://en.wikipedia.org/wiki/Wikipedia:Bypass_your_cache">shift-reload</a>, which will
+  bypass the service worker and serve the page directly from the network.
+</p>
+
+<p>
+  Another remediation is simply to update your service worker script to resolve the bug, and
+  redeploy it to your web server. As soon as the service worker script is expired from a
+  browser's HTTP cache, the updated service worker script will end up being registered. You
+  can use the <a href="../immediate-control/">immediate control</a> pattern in your updated script
+  to ensure that the newly registered service worker takes control over existing tabs. Even if
+  you've inadvertently set the HTTP cache control headers to expire at some point in the far future,
+  a cached copy of the service worker script older than
+  <a href="https://github.com/slightlyoff/ServiceWorker/issues/514">24 hours</a> is automatically
+  ignored, and a call to <code>navigator.serviceWorker.register()</code> will fetch a fresh copy
+  from the network at that point. So the worst-case scenario is that users might have to wait a day
+  until their previously cached service worker script expires.
+</p>
+
+<p>
+  That being said, overly cautious developers might want to add in "kill switch" functionality
+  to their service worker, just in case. One approach to doing this is to call
+  <code>addEventListener('fetch', ...)</code> as early as possible in your service worker script,
+  and register a <code>fetch</code> event listener that will check the request for a special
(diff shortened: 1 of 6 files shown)


service-worker/kill-switch/index.html
```diff
@@ -0,0 +1,61 @@
...
+  you've inadvertently set the HTTP cache control headers to expire at some point in the far future,
+  a cached copy of the service worker script older than
+  <a href="https://github.com/slightlyoff/ServiceWorker/issues/514">24 hours</a> is automatically
+  ignored, and a call to <code>navigator.serviceWorker.register()</code> will fetch a fresh copy
+  from the network at that point. So the worst-case scenario is that users might have to wait a day
+  until their previously cached service worker script expires.
+</p>
+
+<p>
+  That being said, overly cautious developers might want to add in "kill switch" functionality
```
wibblymat, line 40: I might change the word 'overly' to something like 'particularly'. The current wording basically says that you _never_ need to do this, and if you want to do it anyway then you are just being paranoid.

wibblymat:
Actually, I'm not sure if this is useful. In this contrived example the only page that gets returned has instructions for activating the kill switch, but in a real world example it seems unlikely. You only need it when the fetch handler is somehow _broken_, after all.

~100% of the time an affected user would not be able to use this to fix a broken site.

jeffposnick:
I saw http://stackoverflow.com/questions/33986976 and figured that rather than replying directly there, it would be more scalable to put together an "official" sample.

I doubt this is code that I'd ever end up wanting to deploy in my own web apps, but if there's demand for such a thing, maybe it makes sense to show off an implementation that we're confident follows best practices.

Or we could just abandon this PR. Happy to do that too.

wibblymat:
I vote kill it, but would defer to the others if they have an opinion.

service-worker/kill-switch/index.html
```diff
@@ -0,0 +1,61 @@
...
+  service worker that has an undesirable <code>fetch</code> handler, which unconditionally
+  returns useless HTML in response to every fetch request. This simulates the "worst case scenario"
+  for a buggy or poorly implemented service worker. <strong>Once you reload this page, you'll see the
+  undesirable service worker in action.</strong>
+</p>
+
+<p>
+  Before getting into the technical details of the kill switch, it's worth pointing out that even
+  in this "worst case scenario", browsers offer built-in remediations that might prove sufficient.
+  One remediation is to
```
jpmedley, line 20: Add the verb 'press' before shift-reload. Clarifies a little.

service-worker/kill-switch/index.html
```diff
@@ -0,0 +1,61 @@
...
+  <a href="https://en.wikipedia.org/wiki/Wikipedia:Bypass_your_cache">shift-reload</a>, which will
+  bypass the service worker and serve the page directly from the network.
+</p>
+
+<p>
+  Another remediation is simply to update your service worker script to resolve the bug, and
+  redeploy it to your web server. As soon as the service worker script is expired from a
+  browser's HTTP cache, the updated service worker script will end up being registered. You
```
jpmedley, line 28: replace 'end up being' with 'will be'.

service-worker/kill-switch/index.html
jpmedley, line 40: Delete the word 'in'.

service-worker/kill-switch/index.html
```diff
@@ -0,0 +1,61 @@
...
+  ignored, and a call to <code>navigator.serviceWorker.register()</code> will fetch a fresh copy
+  from the network at that point. So the worst-case scenario is that users might have to wait a day
+  until their previously cached service worker script expires.
+</p>
+
+<p>
+  That being said, overly cautious developers might want to add in "kill switch" functionality
+  to their service worker, just in case. One approach to doing this is to call
+  <code>addEventListener('fetch', ...)</code> as early as possible in your service worker script,
+  and register a <code>fetch</code> event listener that will check the request for a special
```
jpmedley, line 43: Replace 'the request' with 'requests'.

service-worker/kill-switch/index.html
```diff
@@ -0,0 +1,61 @@
...
+  URL pattern. If the request URL matches, it will call <code>event.respondWith()</code> to handle
+  the event (preventing any other <code>fetch</code> event listeners from executing), and
+  use <code>registration.unregister()</code> to deactivate itself.
+</p>
+
+<p>
+  At this point, the service worker will not longer control the pages under its previous scope,
```
jpmedley, line 50: Change 'not' to 'no'

service-worker/kill-switch/index.html
```diff
@@ -0,0 +1,61 @@
...
+  but you need to be careful that the buggy service worker doesn't just get registered again.
+  For the purposes of this demo, this page will re-register the same service worker each time you
```
jpmedley, line 52: Replace 'For the purposes of this demo' with 'In this demo'.

jeffposnick:
Killing the kill switch.

PR #266, opened 2015-12-04
Updated web bluetooth device properties

beaufortfrancois:
This CL removes the `paired` property as suggested in https://github.com/WebBluetoothCG/web-bluetooth/issues/189 and adds missing device properties.


jeffposnick:
:+1:

PR #267, opened 2015-12-11
ESLint, with the Google config

### LINKED ISSUE #265: Switch to google ESLint config
Move from JSHint to ESLint, and use https://github.com/google/eslint-config-google as the base config.

Now that the new template setup makes it easy to store our JavaScript in external files, this should be straightforward.

jeffposnick:
R: @gauntface @ebidel @beaufortfrancois @addyosmani @wibblymat or whomever wants to take the time to read through this...

This switches over our linting to ESLint, using the new `eslint-config-google` config as the base.

The majority of the changes boil down to the 80 character line length limit, and missing spaces. I don't think there's anything particularly controversial.

I ended up putting in directory-level `.eslintrc` files in a couple of places when there were specific rules that needed to be specifically ignored in order to properly demonstrate the functionality, or when there's, e.g., code that relies on the `serviceworker` environment. I opted for those instead of inline ESLint config via comments so as not to add extra noise to the code that gets included in the sample landing pages.

Note that we're only linting standalone `.js` files now, so the process of migrating all the inline JavaScript code out of HTML files needs to be completed before we'll have 100% coverage. Given that all new development will have standalone `.js` files, I think we're safe to move forward.

Closes #265


wibblymat:
:+1:

PR #270, opened 2016-01-04
Added Case-insensitive Attribute Selector Matching sample

beaufortfrancois:
Here's a simple PR for the Case-insensitive Attribute Selector Matching upcoming feature in Chrome 49.

R=@jeffposnick

screenshot 2016-01-04 at 2 56 04 pm

### DIFF
--- a/css-attribute-case-sensitivity/index.html
+++ b/css-attribute-case-sensitivity/index.html
@@ -1,0 +1,37 @@
+---
+feature_name: Case-insensitive Attribute Selector Matching
+chrome_version: 49
+feature_id: 5610936115134464
+---
+
+<h3>Background</h3>
+
+<p>By default, case-sensitivity of attribute names and values in selectors
+depends on the document language. That's why an additional modifier (<code>i</code>) for <a
+href="https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors">CSS
+attribute selectors</a> has been added to allow an author to match an
+attribute's value case-insensitively within the ASCII range.</p>
+
+{% capture css %}
+/* All list items which have an id attribute with the ending `case`. */
+li[id$="case" i] {
+  color: green;
+}
+/* All list items which have an id attribute with the exact ending `case` but
+ * not the value endings `CASE` or `Case`. */
+li[id$="case"] {
+  font-weight: bold;
+}
+{% endcapture %}
+
+{% include css_snippet.html css=css %}
+{% capture html %}
+<ul>
+  <li id="foo">foo</li>
+  <li id="snake_case">snake_case</li>
+  <li id="camelCase">camelCase</li>
+  <li id="PascalCase">PascalCase</li>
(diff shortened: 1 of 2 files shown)


css-attribute-case-sensitivity/index.html
```diff
@@ -0,0 +1,36 @@
+---
+feature_name: Case-insensitive Attribute Selector Matching
+chrome_version: 49
+feature_id: 5610936115134464
+---
+
+<h3>Background</h3>
+<p>To allow an author to match a CSS attribute (ASCII) case-insensitively, an
```
jeffposnick: Isn't this a match against an HTML attribute, not a CSS attribute? Also, what's with the `(ASCII)` parenthetical aside? I was [under the impression](https://mathiasbynens.be/notes/html5-id-class) that HTML attributes could contain Unicode characters.

css-attribute-case-sensitivity/index.html
```diff
@@ -0,0 +1,36 @@
...
+additional modifier (<code>i</code>) for CSS attribute selectors has been
```
jeffposnick: Link to https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

css-attribute-case-sensitivity/index.html
```diff
@@ -0,0 +1,36 @@
...
+added.</p>
+<p>By default, case-sensitivity of attribute names and values in selectors
```
jeffposnick, line 9: Where is this documented?

css-attribute-case-sensitivity/index.html
```diff
@@ -0,0 +1,36 @@
...
+ * not the value endings `PDF` or `pDf`. */
+a[download$="pdf"] { 
+  font-weight: bold;
+}
+{% endcapture %}
+
+{% include css_snippet.html css=css %}
+{% capture html %}
+<p>
+  <a href="#" download="foo.pdf">foo.pdf</a>
```
jeffposnick: Could you rewrite this example to use an HTML attribute like `name` that doesn't have user-visible side effects, instead of `download`? There's no benefit to triggering downloads of non-existent files whenever someone clicks on one of these links.

css-attribute-case-sensitivity/index.html
beaufortfrancois, line 9: https://drafts.csswg.org/selectors-4/#attribute-case

beaufortfrancois:
I've reworded the background information and used `id` instead of `download` attribute.
@jeffposnick Is that better?

beaufortfrancois:
Don't merge yet.
I'll update this tomorrow as I'm not completely satisfied with this new patch.

beaufortfrancois:
@jeffposnick Hopefully this patch is better now. 

screenshot 2016-01-05 at 12 11 14 pm

jeffposnick:
:+1:

PR #272, opened 2016-01-04
Fix media query checking for absence of any fine pointer

### LINKED ISSUE #90: media-hover-pointer sample
in the demo https://googlechrome.github.io/samples/media-hover-pointer/ it reads

>  if a user is on a tablet with a connected a mouse, 'any-pointer' will be satisfied. If the mouse is then removed, the browser will report 'any-pointer' as none.

Maybe I'm misunderstanding the spec, but why would/should `any-pointer` be `none` once the mouse is removed? wouldn't the fact that the touchscreen interface is still present count as a pointer device?

patrickhlauke:
As per https://github.com/GoogleChrome/samples/issues/90#issuecomment-85559052 `not (any-pointer: fine)` does not seem to be evaluated correctly in browsers. however, the more verbose `not all and (any-pointer: fine)` works fine (can be verified by using desktop chrome with mouse (small button) and then turning on devtools device emulation (which fails the `any-pointer: fine` media feature test, so the larger font size per this media query kicks in)

Closes #90

### DIFF
--- a/media-hover-pointer/index.html
+++ b/media-hover-pointer/index.html
@@ -66,7 +66,7 @@ limitations under the License.
       .sized {
         font-size: 0.5em;
       }
-      @media not (any-pointer: fine) {
+      @media not all and (any-pointer: fine) {
         .sized {
           font-size: 2.0em;
         }


patrickhlauke:
I signed it!

jeffposnick:
Thanks, @patrickhlauke!

@samthor, I believe this was your sample originally, so I'll defer to you.

I haven't experimented with this functionality very much, but if @samthor can't review, I'll give it a quick once-over.

samthor:
LGTM

patrickhlauke:
:+1:

PR #273, opened 2016-01-07
fix a few typos

zeke:
:wave:

Found this page via http://blog.atom.io/2016/01/07/latest-v8-chromium-features.html


zeke:
I signed it!

beaufortfrancois:
Thank you Zeke!

PR #275, opened 2016-01-14
Add ES2015 Proxies sample

addyosmani:
r: @jeffposnick


addyosmani:
Linting errors I shall fix:

```
   6:21  error  Infix operators must be spaced      space-infix-ops
   6:22  error  Strings must use singlequote        quotes
  14:2   error  More than 1 blank line not allowed  no-multiple-empty-lines
  20:5   error  "handler" is already defined        no-redeclare
  28:26  error  A space is required after ','       comma-spacing
  30:2   error  More than 1 blank line not allowed  no-multiple-empty-lines
```

proxies-es6/demo.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+// Proxying a normal object
+var target = {};
+var handler = {
+  get(receiver, name = 'world') {
```
beaufortfrancois: I'm not sure what is the value of adding a default value to the `name` parameter there.
How can we even trigger it?

proxies-es6/demo.js
```diff
@@ -0,0 +1,43 @@
...
+ChromeSamples.log(p.folks);
+
+// Proxying a function object
+function sum(a, b) {
+  return a + b;
+}
+
+var handler2 = {
+  apply: function(target, thisArg, argumentsList) {
+    console.log(`Calculate sum: ${argumentsList}`);
```
jeffposnick: This line isn't being logged when I test in Chrome Canary 49.0.2622.0:

image

proxies-es6/demo.js
jeffposnick: In addition to @beaufortfrancois's comment, can you double-check the `get` method signature? As per [MSDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/get#Syntax), it's

``` js
var p = new Proxy(target, {
  get: function(target, property, receiver) {
  }
});
```

You're using the [method definition shorthand](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions), but I don't think that should have any effect on the parameter orders, unless there's something implicit going on?

I also think using `property` instead of `name` for that parameter makes things clearer.

proxies-es6/demo.js
```diff
@@ -0,0 +1,43 @@
...
+// Proxying a normal object
+var target = {};
+var handler = {
+  get(receiver, name = 'world') {
+    return `Hello, ${name}!`;
+  }
+};
+
+var p = new Proxy(target, handler);
+ChromeSamples.log(p.folks);
```
jeffposnick: Can you add a comment explicitly stating what this illustrates? Something like
`// p.folks would normally be undefined, but the proxied get handler provides a value.`

proxies-es6/demo.js
```diff
@@ -0,0 +1,43 @@
...
+
+// Proxying a function object
+function sum(a, b) {
+  return a + b;
+}
+
+var handler2 = {
+  apply: function(target, thisArg, argumentsList) {
```
jeffposnick: You're using [method definition shorthand](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions) in the previous handler and `function` here.

I'd recommend consistently using one or the other.

proxies-es6/demo.js
```diff
@@ -0,0 +1,43 @@
...
+    console.log(`Calculate sum: ${argumentsList}`);
+    return target.apply(thisArg, argumentsList);
+  }
+};
+
+var proxy = new Proxy(sum, handler2);
```
jeffposnick: You've got `handler` and `handler2` as variable names. It could be clearer to use `proxy` and `proxy2` for the corresponding proxies.

proxies-es6/demo.js
```diff
@@ -0,0 +1,43 @@
...
+ChromeSamples.log(proxy(1, 2));
+
+// Field interception with Proxy and the Reflect API
```
jeffposnick: What's the Reflect API? Can you link to some relevant docs?

I see that this example differs from the first in that you're setting properties on the `Proxy` instead of a normal `Object`, but I don't understand the motivation for that. Maybe explaining something about the Reflect API would clear that up?

proxies-es6/demo.js
```diff
@@ -0,0 +1,43 @@
...
+var pioneer = new Proxy({}, {
+  get: function(target, name, receiver) {
```
jeffposnick: This `get` doesn't get exercised in your example.

jeffposnick:
Reviewed. Thanks for writing this!

jeffposnick:
Hey @addyosmani—what are your plans for this PR, given that revised versions of most of this code is now live at https://developers.google.com/web/updates/2016/02/es2015-proxies?

jeffposnick:
Friendly **ping** re: whether we can close this out.

addyosmani:
I'll clean this up this week and repush. Sorry for the delay again!

addyosmani:
@jeffposnick Pushed a new commit attempting to address feedback. PTAL when you get a chance. Thanks!

proxies-es6/demo.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+// Proxying a normal object
+var target = {};
+var handler = {
+  get(target, property, receiver) {
```
jeffposnick: ESLint fails with

```
/home/travis/build/GoogleChrome/samples/proxies-es6/demo.js
  6:25  error  "receiver" is defined but never used  no-unused-vars
```

I see the benefit of including `receiver` as a parameter so that developers know it'll get passed in, so if you're not going to make use of it in the method body, could you add in a `// eslint-disable-line no-unused-vars` at the end of the line?

jeffposnick:
I took the liberty of fixing the linting error and merging 😄

PR #276, opened 2016-01-15
Check for request.mode === 'navigate'

jeffposnick:

Support for a Fetch API request mode of `'navigate'` was added in M49, and this updates an existing service worker sample to make use of it.


beaufortfrancois:
:+1:

PR #279, opened 2016-01-20
High Resolution Time Stamp for Events Sample

jeffposnick:
R: @addyosmani @ebidel @beaufortfrancois @gauntface etc.
CC: @RByers

Here's the example for the new [High Resolution Time Stamp for Events](https://www.chromestatus.com/feature/5523910145605632) functionality in M49.

There's a live version deployed at https://jeffy.info/samples/event-timestamp/

I'd love to make sure @RByers doesn't have any issues with how I'm representing the information and example prior to merging.

### DIFF
--- a/event-timestamp/demo.js
+++ b/event-timestamp/demo.js
@@ -1,28 +1,25 @@
-var dateNowAtLoad = Date.now();
-var offsetFromEpoch;
-
+var offsetFromEpoch = 0;
 var previousX = 0;
 var previousY = 0;
 var previousT = 0;
 
-function calculateDateForTimestamp(timestamp) {
-  if (offsetFromEpoch === undefined) {
-    // When we don't yet know what the offset should be, use a simple heuristic:
-    // If the timestamp value is less than the time this JS was first loaded,
-    // then it's almost certainly a DOMHighResTimeStamp. Adding timestamp to
-    // performance.timing.navigationStart should give a value relative to
-    // the epoch, which could be passed in to the Date constructor.
-    // Otherwise, it's a DOMTimeStamp, and is already relative to the epoch.
-    offsetFromEpoch = (timestamp < dateNowAtLoad) ?
-      performance.timing.navigationStart : 0;
+window.addEventListener('load', function(event) {
+  // Use a simple heuristic to determine the offset from the epoch:
+  // If the timestamp value is less than the current time,
+  // then it's almost certainly a DOMHighResTimeStamp. Adding timestamp to
+  // performance.timing.navigationStart should give a value relative to
+  // the epoch, which could be passed in to the Date constructor.
+  // Otherwise, it's a DOMTimeStamp, and is already relative to the epoch.
+  if (event.timeStamp < Date.now()) {
+    offsetFromEpoch = performance.timing.navigationStart;
   }
 
-  return new Date(offsetFromEpoch + timestamp);
-}
+  // While a 'mousemove' listener is being used for the purposes of this
+  // example, the new timeStamp resolution applies to all Event types.
+  window.addEventListener('mousemove', calculateVelocity);
+});
 
-// While a 'mousemove' listener is being used for the purposes of this
-// example, the new timeStamp resolution applies to all Event types.
-window.addEventListener('mousemove', function(event) {
+function calculateVelocity(event) {
   var Δx = event.screenX - previousX;
   var Δy = event.screenY - previousY;
   var Δd = Math.sqrt(Math.pow(Δx, 2) + Math.pow(Δy, 2));
@@ -37,8 +34,9 @@ window.addEventListener('mousemove', function(event) {
   // Multiply by 1000 to go from milliseconds to seconds.
   var velocityInPixelsPerSecond = Δd / Δt * 1000;
 
-  // Contrived example to illustrate how to convert event.timeStamp to a Date.
-  var date = calculateDateForTimestamp(event.timeStamp);
+  // Contrived example to illustrate how to use offsetFromEpoch to get a Date
+  // that corresponds to event.timeStamp.
+  var date = new Date(event.timeStamp + offsetFromEpoch);
 
   ChromeSamples.setStatus(velocityInPixelsPerSecond + ' as of ' +
     date.toLocaleTimeString());
@@ -46,4 +44,4 @@ window.addEventListener('mousemove', function(event) {
   previousX = event.screenX;
   previousY = event.screenY;
   previousT = event.timeStamp;
-});
+}
--- a/event-timestamp/demo.js
+++ b/event-timestamp/demo.js
@@ -1,6 +1,24 @@
+var dateNowAtLoad = Date.now();
+var offsetFromEpoch;
+
 var previousX = 0;
 var previousY = 0;
 var previousT = 0;
+
+function calculateDateForTimestamp(timestamp) {
+  if (offsetFromEpoch === undefined) {
+    // When we don't yet know what the offset should be, use a simple heuristic:
+    // If the timestamp value is less than the time this JS was first loaded,
+    // then it's almost certainly a DOMHighResTimeStamp. Adding timestamp to
+    // performance.timing.navigationStart should give a value relative to
+    // the epoch, which could be passed in to the Date constructor.
+    // Otherwise, it's a DOMTimeStamp, and is already relative to the epoch.
+    offsetFromEpoch = (timestamp < dateNowAtLoad) ?
+      performance.timing.navigationStart : 0;
+  }
+
+  return new Date(offsetFromEpoch + timestamp);
+}
 
 // While a 'mousemove' listener is being used for the purposes of this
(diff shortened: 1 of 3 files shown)


RByers:
Looking now.  + @majido who owns this feature.  Did you see my sample code in http://rbyers.github.io/scroll-latency.html?

event-timestamp/demo.js
```diff
@@ -0,0 +1,27 @@
+var previousX = 0;
+var previousY = 0;
+var previousT = 0;
+
+// While a 'mousemove' listener is being used for the purposes of this
+// example, the new timeStamp resolution applies to all Event types.
+window.addEventListener('mousemove', function(event) {
+  var Δx = event.screenX - previousX;
```
ebidel: `var Δx` is cute, but should we really use it in a sample. When I originally saw the demo I thought it was an encoding error :)

ebidel:
Just the comment from me.

RByers:
Sample looks great to me, thanks!  A couple small comments:

1) This sample neatly side-steps worrying about feature detection (eg. in case you really do want to compare against wall time).  In practice the range of values is so different that it's pretty easy to detect the feature by comparing values to performance.now and Date.now (see checkTimebase [here](http://rbyers.github.io/scroll-latency.js)).  Is that worth showing here, or better to keep it simple?

2) The real reason we're making this change is as part of exposing the true system time an input event was generated.  I don't _think_ we want to go into any detail in the sample here (we're still not sure we really want people to use this as a [poor-mans input-latency metric](rbyers.github.io/scroll-latency.html)).  I tried to come up with something small to say about this, but the more I think about it the more it seems simpler / better to leave the sample as is and not mention this.  Accurate velocity measurement is a great use case though!

majido:
> 1) This sample neatly side-steps worrying about feature detection (eg. in case you really do want to compare against wall time). In practice the range of values is so different that it's pretty easy to detect the feature by comparing values to performance.now and Date.now (see checkTimebase here).  Is that worth showing here, or better to keep it simple?

The sample indeed sidesteps that issue but the description suggests a method to convert a DOMHighResTimeStamp to a DOMTimeStamp which can break if used naively without detecting the timebase. I would rather legacy proof that code by adding the timebase detection or instead add a clear warning about it. 

@RByers code for timebase detection is good but I think it can be made simpler and more efficient but assuming event.timeStamp is consistent across all events which is a valid assumption.

event-timestamp/demo.js
jeffposnick: Why are you conspiring to deprive me of the little joy Unicode has to offer?

ebidel: :) keep it if you like it

jeffposnick:
I could definitely see the base value for `timeStamp` changing leading to developer confusion/breakage, so I'm fine with including a concrete example of detecting the proper base as part of this sample.

(We'll also need to explicitly communicate this breaking change via various other channels.)

I've pushed an update that includes the following logic, which _seems_ safe, but I'm happy to adjust as needed:

``` js
var dateNowAtLoad = Date.now();
var offsetFromEpoch;

function calculateDateForTimestamp(timestamp) {
  if (offsetFromEpoch === undefined) {
    // When we don't yet know what the offset should be, use a simple heuristic:
    // If the timestamp value is less than the time this JS was first loaded,
    // then it's almost certainly a DOMHighResTimeStamp. Adding timestamp to
    // performance.timing.navigationStart should give a value relative to
    // the epoch, which could be passed in to the Date constructor.
    // Otherwise, it's a DOMTimeStamp, and is already relative to the epoch.
    offsetFromEpoch = (timestamp < dateNowAtLoad) ?
      performance.timing.navigationStart : 0;
  }

  return new Date(offsetFromEpoch + timestamp);
}
```

jeffposnick:
After thinking about this some more, I've modified the `offsetFromEpoch` to take place during an initial `window.onload` handler. That should prevent against the off chance that someone changes their system clock time in between when the page is first loaded and when they first move their mouse. It also makes the code a little cleaner, since I can delay setting up the `window.onmousemove` handler until we're sure `offsetFromEpoch` has been calculated.

I'm interested to hear what folks think about this approach. I'm including it inline for easy reference.

``` js
window.addEventListener('load', function(event) {
  // Use a simple heuristic to determine the offset from the epoch:
  // If the timestamp value is less than the current time,
  // then it's almost certainly a DOMHighResTimeStamp. Adding timestamp to
  // performance.timing.navigationStart should give a value relative to
  // the epoch, which could be passed in to the Date constructor.
  // Otherwise, it's a DOMTimeStamp, and is already relative to the epoch.
  if (event.timeStamp < Date.now()) {
    offsetFromEpoch = performance.timing.navigationStart;
  }

  // While a 'mousemove' listener is being used for the purposes of this
  // example, the new timeStamp resolution applies to all Event types.
  window.addEventListener('mousemove', calculateVelocity);
});
```

RByers:
I don't think we want to encourage people to compare `Event.timeStamp` values to Date.now().  Date.now() is flawed for most such use-cases because it's wall time - suject to NTP skew, users changing their clock etc.

Inverting the logic to focus on monotonic timestamps (such as DOMHighRestTimeStamp) would be better.  Eg: maybe something like:

``` javascript
window.addEventListener('load', function(event) {
  // Use a simple heuristic to support browsers which still report event timestamps relative
  // to the UNIX epoch, instead of relative to navigation start.
  if (event.timeStamp > performance.now() + 1000) {
    offsetToTimeOrigin = performance.timing.navigationStart;
  }

  // While a 'mousemove' listener is being used for the purposes of this
  // example, the new timeStamp resolution applies to all Event types.
  window.addEventListener('mousemove', calculateVelocity);
});  
....
var eventTime = event.timeStamp - offsetToTimeOrigin;
var eventLatency = performance.now() - eventTime;
```

@majido is also working on a [polyfill](https://gist.github.com/majido/e2d33d1232b07a79ce29) to make other browsers conform to the new spec.

jeffposnick:
For new code, I'm sure developers might want to make the adjustment you're proposing, so that they end up with a value that would be easy to compare to `performance.now()`, even in legacy browsers. And comparing against `performance.now()` has the other benefits you describe.

But I feel like providing an example of how to convert in the other direction would best help developers who would otherwise be caught unawares by this breaking change. I'm assuming there are developers with legacy code that expects `Event.timeStamp` to be offset from the epoch, and we can help that code continue working in M49 (without a too much of a rewrite) by showing them how to convert to an equivalent value.

We can link to multiple samples from the Chrome Status entry, so http://rbyers.github.io/scroll-latency.html will remain there too.

What do you all think? We can structure things a bit differently, but I want to do something to ease developers through the breaking change.

majido:
I agree that there is value to provide a easy to use fix. Note that such a code is _already_ broken to some extent in Firefox because Firefox `Event.timeStamp` may use a different timebase based on the event type, it can be since epoch (custom events), since last reboot (most input event), or 0 (load, animation events) so naively comparing it to Date.now() is already broken to some extent.

Your snippet that relies on load event timestamp actually does not work because getting Date.now() in the event handler is certainly going to be larger or equal the event timestamp even for legacy time stamps. (FYI, Firefox load event timestamp is 0 :cry:  )

I think a batter way to do detect this is by creating a custom event and checking its timestamp against Date.now(). This still does not work in FF at least for input event. Here is a sample:

``` js
toDOMTimeStamp = (function(){
  var now = Date.now();
  var testEventTimeStamp = document.createEvent('KeyboardEvent').timeStamp;

  var offsetFromEpoch = 0;
  if (testEventTimeStamp < now ) {
    // The browser is using DOMHighResTimeStamp so adjust offset. 
    offsetFromEpoch = performance.timing.navigationStart;
  }

  return function(timeStamp) {
    return Math.round(timeStamp + offsetFromEpoch);
  }
})();

```

RByers:
Yeah I get the argument that we should be primarily concerned with helping developers migrate existing code.  Should we just provide [the polyfill](https://gist.github.com/majido/e2d33d1232b07a79ce29) as a sample for that, and ignore all this from the other sample(s)?

Anyway I don't think this is worth agonizing over - we know it's rare for anyone to compare Event.timeStamp to anything other than another Event.timeStamp.  So I'm happy either way.

jeffposnick:
I'm sorry, I didn't realize that Firefox's current `Event.timeStamp` was already different from pre-M49 Chrome's. I guess that means that developers aren't blindly doing things like passing `Event.timeStamp` to `new Date()`. Thanks for the additional context, and that makes me less concerned than I was about the change in behavior.

Given that, I'm just going to remove the portion of the sample that attempts to coerce the timestamp into a value that could be passed to `new Date()`, and just focus on the change in timestamp value in-between event handler invocations.

jeffposnick:
Okay, I've simplified the sample as described above. It should be similar to what was initially reviewed, so I'll merge this.

I'm also on the hook to write up something for https://developers.google.com/web/updates/ explaining this change, and I'll use that opportunity to elaboate on the backwards compatability implications and polyfill recommendations.

PR #280, opened 2016-01-21
bluetooth: a DataView is returned instead of an ArrayBuffer in M50

beaufortfrancois:
R=@jeffposnick
TBR=@g-ortuno

### DIFF
--- a/web-bluetooth/battery-level.js
+++ b/web-bluetooth/battery-level.js
@@ -21,9 +21,10 @@ function onButtonClick() {
     log('Reading Battery Level...');
     return characteristic.readValue();
   })
-  .then(buffer => {
-    let data = new DataView(buffer);
-    let batteryLevel = data.getUint8(0);
+  .then(value => {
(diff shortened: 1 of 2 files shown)


web-bluetooth/battery-level.js
```diff
@@ -21,9 +21,10 @@ function onButtonClick() {
     log('Reading Battery Level...');
     return characteristic.readValue();
   })
-  .then(buffer => {
-    let data = new DataView(buffer);
-    let batteryLevel = data.getUint8(0);
+  .then(value => {
+    // In Chrome 50+, a DataView is returned instead of an ArrayBuffer.
```
g-ortuno, line 25: We should set a reminder to remove this once 50 hits stable.

g-ortuno:
lgtm from API point of view

web-bluetooth/battery-level.js
beaufortfrancois, line 25: Already added one in Keep ;)

PR #281, opened 2016-01-21
CSS Custom Properties sample

sgomes:
Adds a CSS Custom Properties sample that illustrates basic usage, cascading, and more advanced usage with `calc()`.

@jeffposnick PTAL!


jeffposnick:
CC: @shans, whom I believe is the Chromium engineer who worked on the functionality, as a heads-up. (Apologies if I've tagged the wrong GitHub account.)

sgomes:
For easier testing, here's the GitHub page for this sample on my branch: https://sgomes.github.io/samples/css-custom-properties/

css-custom-properties/index.css
```diff
@@ -0,0 +1,186 @@
...
+  --grid-gutter: (var(--margins) * var(--spacing-unit));
+  /* How big the space should be around the grid */
+  --grid-margin: (var(--margin-size) * var(--spacing-unit));
+  /* Calculated cell margin */
+  --cell-margin: (var(--grid-gutter) / 2);
+}
+
+
+body {
+  /* Resets */
```
jeffposnick: Could you exclude this? We've got a consistent style for `<body>` defined at https://github.com/GoogleChrome/samples/blob/gh-pages/styles/main.css#L1

css-custom-properties/demo.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  // Retrieve initial custom property values and update controls.
+  quantum.value = getVariable(styles, '--spacing-unit').replace('px', '');
+  gutter.value = getVariable(styles, '--margins');
+  columns.value = getVariable(styles, '--grid-columns');
+
+  // Set up event handlers for having the sliders update the custom properties
+  // at the document level.
+  quantum.addEventListener('change', function() {
```
jeffposnick: How would you feel about triggering these callbacks on `input` instead of `change`? It's easier to understand what's going on if you can drag the slider and see changes instantly.

css-custom-properties/index.html
```diff
@@ -0,0 +1,200 @@
...
+      </button>
+    </div>
+  </div>
+</div>
+{% endcapture %}
+{% include output_helper.html initial_output_content=initial_output_content %}
+
+<h3>CSS Snippet</h3>
+
+{% highlight css %}
```
jeffposnick: Instead of including an extra copy of the CSS inside of the `{% highlight css %}` block, you can use the `css_snippet.html` helper to both display the highlighted CSS inline, as well as include it in the page via a `<style>` tag.

https://github.com/GoogleChrome/samples/blob/gh-pages/css-motion-path/index.html#L59 is an example.

Given that the CSS is so important to this example, I'd recommend doing away with the external `index.css` file entirely, and put all of your CSS inline here using `css_snippet.html`.

jeffposnick:
Reviewed. Thanks for putting this together!

css-custom-properties/demo.js
```diff
@@ -0,0 +1,63 @@
...
+  // Get the values for the button's colours...
+  var primary = getVariable(styles, '--primary-color');
+  var text = getVariable(styles, '--primary-color-text');
+  // ... and apply them to the document.
+  setDocumentVariable('--primary-color', primary);
+  setDocumentVariable('--primary-color-text', text);
+};
+
+// Initialise page controls.
+window.onload = function() {
```
ebidel: `window.addEventListener`

css-custom-properties/demo.js
```diff
@@ -0,0 +1,63 @@
...
+  // Get the styles for the document.
+  // This is where we've chosen to store all the global variables we use.
+  var styles = getComputedStyle(document.documentElement);
+
+  var quantum = document.getElementById('quantum');
+  var gutter = document.getElementById('gutter');
+  var columns = document.getElementById('columns');
+
+  // Set up event handlers for buttons.
+  var buttons = document.getElementsByClassName('picker-button');
```
ebidel: `querySelectorAll('.picker-button')`

ebidel:
May have missed it, but what about an example of default values?

`color: var(--my-toolbar-title-color, blue);`

Theming web components is a great use case.

css-custom-properties/index.css
sgomes: Done.

css-custom-properties/demo.js
sgomes: Done.

css-custom-properties/index.html
sgomes: Done.

css-custom-properties/demo.js
sgomes: Done.

css-custom-properties/demo.js
sgomes: Done.

sgomes:
@ebidel: I set up all the "defaults" by defining the variables at the document level. But I agree with this being a good thing to show, so I've added one of those and a comment to illustrate usage. How does that look?

sgomes:
@jeffposnick and @ebidel: Thanks for the super-speedy review! :) I think I've addressed all your comments, let me know if there's anything I missed!

jeffposnick:
:+1:. I'll defer to @ebidel prior to merging to make sure he's happy.

addyosmani:
LGTM as is. Will also defer to Eric for final review.

ebidel:
Lgtm x 3

addyosmani:
Thanks all. Landed via https://github.com/GoogleChrome/samples/commit/e580321706e00f2a55731dbab7dace4e894df8fe and https://github.com/GoogleChrome/samples/commit/589f5d74c5c35f858314908f9584145bdb6e7165

PR #282, opened 2016-01-21
WindowClient.navigate() sample

jeffposnick:
R: @addyosmani @gauntface @jakearchibald @beaufortfrancois etc.

This is a sample for `WindowClient.navigate()`, shipping in [Chrome 49](https://www.chromestatus.com/feature/5314750808326144).

You can view a live version of this sample at https://jeffy.info/samples/service-worker/windowclient-navigate/

### DIFF
--- a/service-worker/windowclient-navigate/index.html
+++ b/service-worker/windowclient-navigate/index.html
@@ -1,0 +1,31 @@
+---
+feature_name: "Service Worker: WindowClient.navigate()"
+chrome_version: 49
+feature_id: 5314750808326144
+---
+
+<h3>Background</h3>
+<p>
+  <code>WindowClient.navigate()</code> allows a service worker to cause a web
+  page to navigate to a specific URL.
+</p>
+
+<p>
+  This sample shows how <code>WindowClient.navigate()</code> can be called
+  within a service worker's <code>activate</code> handler, to navigate to
+  <code>activated.html</code>. If your browser supports service workers, and
+  supports <code>WindowClient.navigate()</code>, then you'll be automatically
+  navigated away from this page once the service worker activates.
+</p>
+
+<p>
+  Note that the <code>activate</code> event is only fired once per version of
+  a given service worker registration. If you'd like to trigger it again to
+  test out the <code>WindowClient.navigate()</code> behavior, you can revisit
+  this page in a Chrome
+  <a href="https://support.google.com/chrome/answer/95464?source=gsearch&hl=en">Incognito window</a>, which will trigger a fresh
+  service worker registration.
+</p>
+
+{% include js_snippet.html filename='index.js' title="This Page's JavaScript" %}
+{% include js_snippet.html filename='service-worker.js' displayonly=true title="Service Worker's JavaScript" %}
--- a/service-worker/windowclient-navigate/index.html
+++ b/service-worker/windowclient-navigate/index.html
@@ -6,7 +6,7 @@ feature_id: 5314750808326144
 
 <h3>Background</h3>
 <p>
-  <code>WindowClient.navigate()</code> allows a service worker to cause a web
+  <code><a href="https://developer.mozilla.org/en-US/docs/Web/API/WindowClient">WindowClient</a>.navigate()</code> allows a service worker to cause a web
   page to navigate to a specific URL.
 </p>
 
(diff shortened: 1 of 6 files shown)


service-worker/windowclient-navigate/index.html
```diff
@@ -0,0 +1,31 @@
...
+  supports <code>WindowClient.navigate()</code>, then you'll be automatically
+  navigated away from this page once the service worker activates.
+</p>
+
+<p>
+  Note that the <code>activate</code> event is only fired once per version of
+  a given service worker registration. If you'd like to trigger it again to
+  test out the <code>WindowClient.navigate()</code> behavior, you can revisit
+  this page in a Chrome
+  <a href="https://support.google.com/chrome/answer/95464?source=gsearch&hl=en">Incognito window</a>, which will trigger a fresh
```
beaufortfrancois, line 26: We may want to mention DevTools as well, as it's possible to delete a service worker from there and refresh the page.
screenshot 2016-01-22 at 8 16 41 am

service-worker/windowclient-navigate/index.html
```diff
@@ -0,0 +1,31 @@
+---
+feature_name: "Service Worker: WindowClient.navigate()"
+chrome_version: 49
+feature_id: 5314750808326144
+---
+
+<h3>Background</h3>
+<p>
+  <code>WindowClient.navigate()</code> allows a service worker to cause a web
```
beaufortfrancois: @jpmedley It might be good to document <code>WindowClient.navigate()</code> on MDN and link from there.  WDYT?

service-worker/windowclient-navigate/index.html
```diff
@@ -0,0 +1,31 @@
...
+  page to navigate to a specific URL.
+</p>
+
+<p>
+  This sample shows how <code>WindowClient.navigate()</code> can be called
+  within a service worker's <code>activate</code> handler, to navigate to
+  <code>activated.html</code>. If your browser supports service workers, and
+  supports <code>WindowClient.navigate()</code>, then you'll be automatically
+  navigated away from this page once the service worker activates.
```
beaufortfrancois, line 18: We may want to mention that you cannot _yet_ navigate to out-of-scope URLs.

beaufortfrancois:
LGTM with nits.

service-worker/windowclient-navigate/index.html
jeffposnick, line 18: That doesn't appear to be true. I switched the SW code to `return client.navigate('https://www.google.com/');` and it worked.

service-worker/windowclient-navigate/index.html
jeffposnick: I can just add [`WindowClient`](https://developer.mozilla.org/en-US/docs/Web/API/WindowClient) in the interim, as that should direct folks to the right place.

service-worker/windowclient-navigate/index.html
jeffposnick, line 26: If there were a good, standalone link that detailed the current DevTools service worker support I'd do that. As it is, I'd rather just link to the Incognito help center article.

beaufortfrancois, line 26: We could update https://www.chromium.org/blink/serviceworker/service-worker-faq#TOC-SW-debugging-panel-in-Resources section.

service-worker/windowclient-navigate/index.html
beaufortfrancois, line 18: Weird. See https://code.google.com/p/chromium/issues/detail?id=500911#c23 and https://www.w3.org/TR/service-workers/#client-navigate-method

jeffposnick, line 18: For `client.navigate(destination)`, I believe the restriction regarding scope applies to `client`. It needs to be a `WindowClient` that's in scope for the service worker in question. I don't believe that restriction applies to `destination`.

That matches what I'm seeing in practice against M49. It also matches my interpretation of the negative example at https://github.com/slightlyoff/ServiceWorker/issues/752#issue-107657168

beaufortfrancois, line 18: Oops. You're right.
Thanks for taking the time to explain.

PR #284, opened 2016-01-25
Cookie Prefixes sample

jeffposnick:
R: @mikewest @addyosmani @ebidel @beaufortfrancois @gauntface etc.

Here's my sample for [Cookie Prefixes](https://www.chromestatus.com/feature/4952188392570880), shipping in M49.

It would be nice if @mikewest could give a :+1: to the explanation I include, which is mostly cribbed from his spec draft. But I might be misrepresenting something...

There's a live version at https://jeffy.info/samples/cookie-prefixes/

### DIFF
--- a/cookie-prefixes/index.html
+++ b/cookie-prefixes/index.html
@@ -20,9 +20,10 @@ feature_id: 4952188392570880
     is required.
   </li>
   <li>
-    <code>__Host-</code>, which signals to the browser that the
-    <a href="http://tools.ietf.org/html/rfc6265#section-4.1.2.4"><code>Path=/</code> attribute</a>
-    is required, and at the same time, that the <a href="http://tools.ietf.org/html/rfc6265#section-4.1.2.3"><code>Domain</code> attribute</a>
+    <code>__Host-</code>, which signals to the browser that both the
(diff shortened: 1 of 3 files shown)


cookie-prefixes/index.html
```diff
@@ -0,0 +1,47 @@
...
+<ul>
+  <li>
+    <code>__Secure-</code>, which signals to the browser that the
+    <a href="http://tools.ietf.org/html/rfc6265#section-4.1.2.5"><code>Secure</code> attribute</a>
+    is required.
+  </li>
+  <li>
+    <code>__Host-</code>, which signals to the browser that the
+    <a href="http://tools.ietf.org/html/rfc6265#section-4.1.2.4"><code>Path=/</code> attribute</a>
+    is required, and at the same time, that the <a href="http://tools.ietf.org/html/rfc6265#section-4.1.2.3"><code>Domain</code> attribute</a>
```
mikewest: `Secure` is required as well.

mikewest:
One nit, otherwise LGTM.

jeffposnick:
Thanks for the review! I'm merging now that the feedback is addressed.

PR #285, opened 2016-01-26
Update sample for consistency

joemarini:
Other samples use addEventListener. Updating this one to be consistent


jeffposnick:
This is @paulirish's sample, from way back when. CC:ing him as an FYI.

PR #287, opened 2016-01-26
Add ES2015 destructuring sample

addyosmani:
r: @jeffposnick

### DIFF
--- a/destructuring-es6/demo.js
+++ b/destructuring-es6/demo.js
@@ -1,0 +1,39 @@
+// ES2015 destructuring assignment lets you extract data from arrays or objects
+// using a syntax similar to the construction of array and object literals
+var cats = ['😹', '🙀', '😻'];
+var catAges = {sam: 12, passy: 14, surma: 16};
+
+// Access the first three items of an array in ES5
+var one = cats[0];
+var two = cats[1];
+var three = cats[2];
+ChromeSamples.log(one, two, three);
+
+// Destructure the array in ES2015
+var [a, b, c] = cats;
+ChromeSamples.log(a, b, c);
+
+// Destructure an array omitting certain values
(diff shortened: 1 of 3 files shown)


destructuring-es6/demo.js
```diff
@@ -0,0 +1,39 @@
...
+
+// Destructure an object
+var {passy, surma} = catAges;
+ChromeSamples.log(passy, surma);
+
+// Destructure an object, binding variables to different properties
+var {sam: cat1, passy: cat2, surma: cat3} = catAges;
+ChromeSamples.log(cat1, cat2, cat3);
+
+// Default values for function object parameters
```
jeffposnick, line 32: Could you add a link to

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Destructured_parameter_with_default_value_assignment

in the comment to provide some context about default values for parameters? It's technically distinct from destructuring and folks might not be familiar with it.

destructuring-es6/demo.js
```diff
@@ -0,0 +1,39 @@
...
+var request = ({
```
jeffposnick: Breaking this up due to line length limits ends up hurting readability.

How about using `cors` as a parameter instead of `crossDomain` (since you're just making stuff up anyway)? You could fit everything on one line that way:

``` js
var request = ({url: url = 'github.com', cors: cors = true}, ...data) => {
  ChromeSamples.log(`url: ${url}, cors: ${cors}, data: ${data}`);
};
```

jeffposnick:
Reviewed. Thanks!

addyosmani:
Thanks. PTAL

jeffposnick:
:+1:

PR #289, opened 2016-01-27
Add ES2015 default parameters

addyosmani:
r: @jeffposnick

### DIFF
--- a/default-parameters-es6/demo.js
+++ b/default-parameters-es6/demo.js
@@ -5,7 +5,7 @@ function greetES5(message, to) {
   // Test parameter values in the function body, specifying a default
   to = to || 'DOM';
   // OR if (typeof to === undefined) { to = 'DOM'; }
-  // OR if (arguments.to === 1) { to = DOM'; '}
+  // OR if (arguments.length === 1) { to = 'DOM'; }
   ChromeSamples.log(message + ', ' + to);
 }
 greetES5('Good morning');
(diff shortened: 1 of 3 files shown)


default-parameters-es6/demo.js
```diff
@@ -0,0 +1,41 @@
+// Supply default values for parameters
+
+// Before, using ES5...
+function greetES5(message, to) {
+  // Test parameter values in the function body, specifying a default
+  to = to || 'DOM';
+  // OR if (typeof to === undefined) { to = 'DOM'; }
+  // OR if (arguments.to === 1) { to = DOM'; '}
```
jeffposnick: I don't understand `arguments.to` in this context. Do you mean `arguments.length`?

Also, your `'` character in `to = DOM'; '` seems to have floated off to the wrong position.

default-parameters-es6/demo.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+f(3);
+// For more info, read: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Passing_undefined
+
+// Default arguments are evaluated at call time. This means
+// a new array is created each time the function is called here:
+function append(value, array = []) {
+  array.push(value);
+  return array;
```
jeffposnick: `ChromeSamples.log(array);`

default-parameters-es6/demo.js
```diff
@@ -0,0 +1,41 @@
...
+  ChromeSamples.log(message + ', ' + to);
+}
+
+greet('Good morning');
+greet('Sup', 'CSS');
+
+// Another example
+function f(x, y = 12) {
+  // y is 12 if not passed (or passed as undefined)
+  return x + y;
```
jeffposnick:
``` js
ChromeSamples.log(`${x} + ${y} = ${x + y}`);
```

default-parameters-es6/demo.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+f(3);
```
jeffposnick, line 29: Can you add in an additional example to hammer home how false-y values like `0` won't trigger the default? Something like

``` js
// The default parameter won't be used since we're passing in a value, even though 0 is false-y.
f(5, 0);
```

jeffposnick:
Reviewed. Thanks!

addyosmani:
Attempted to address comments. PTAL. Thanks for reviewing!

jeffposnick:
:+1:

PR #293, opened 2016-02-08
web-bluetooth: specify blacklisted UUIDs

beaufortfrancois:
This CL simply specificies which characterististics/services are now blacklisted from read/write actions.

https://codereview.chromium.org/1649253002

R=@jeffposnick


jeffposnick:
Hey @beaufortfrancois—It's hard to tell what changed as part of the blacklisting, since it looks like there's a diff for every line in all those files due to adding content to the `<option>` elements.

So what's been removed? Is it the sort of thing that needs some explanatory comment in the sample or code so that someone doesn't get confused why a UUID that would otherwise work doesn't?

beaufortfrancois:
I didn't add any characteristics/services there.
I "simply" added a label to each `<option>` and appended "(blacklisted)" for 3 of them.
That's all there ;)

jeffposnick:
Sorry, I missed that from scanning through the diff :smile:

beaufortfrancois:
Thx!

PR #299, opened 2016-02-11
Use page.path, not page.url

jeffposnick:
R: @beaufortfrancois 

This seems to help... see https://jeffy.info/samples/web-bluetooth/device-disconnect.html


beaufortfrancois:
Much better on web-bluetooth samples
:+1:

PR #305, opened 2016-02-24
Adjust default style

beaufortfrancois:
screenshot 2016-02-24 at 3 29 25 pm

R=@jeffposnick


jeffposnick:
Sure, looks a bit neater to me. Thanks!

PR #308, opened 2016-02-29
web-bluetooth: Add service.getCharacteristics sample

beaufortfrancois:
R=@jeffposnick


jeffposnick:
:+1: from the limited perspective I have with the Bluetooth API.

PR #310, opened 2016-03-07
HTMLMediaElement.play() promise interface

jeffposnick:
R: @addyosmani @PaulKinlan @gauntface @beaufortfrancois @ebidel etc.

Here's the code sample for the [HTMLMediaElement.play() returns a promise](https://www.chromestatus.com/feature/5920584248590336) feature in M50.

It makes use of an existing audio file that we have rights for and which was used for another sample.

You can view a live version at https://jeffy.info/samples/play-return-promise/

There's a separate /web/updates/ article planned with more information.

### DIFF
--- a/play-return-promise/README.md
+++ b/play-return-promise/README.md
@@ -1,0 +1,5 @@
+HTMLMediaElement.play() Returns a Promise Sample
+===
+See https://googlechrome.github.io/samples/play-return-promise/index.html for a live demo.
+
+Learn more at https://www.chromestatus.com/feature/5920584248590336
--- a/play-return-promise/demo.js
+++ b/play-return-promise/demo.js
@@ -1,0 +1,16 @@
+function startPlayback() {
+  return document.querySelector('#music').play();
+}
+
+ChromeSamples.log('Attempting to start playback automatically...');
+
+startPlayback().then(function() {
+  ChromeSamples.log('The play() Promise resolved! Rock on!');
+}).catch(function(error) {
+  ChromeSamples.log('The play() Promise rejected!\nUse the Play button instead.\n' + error);
+
+  var playButton = document.querySelector('#play');
+  // The user interaction requirement will be met if playback starts via a click event.
+  playButton.addEventListener('click', startPlayback);
+  playButton.hidden = false;
+});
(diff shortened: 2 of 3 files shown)


beaufortfrancois:
I would simply add the html so that user understands that `#music` is an audio HTML tag.
Moreover, to make it nicer, I'd shorten logs.

See below.
Otherwise :+1: 

``` diff
diff --git a/play-return-promise/demo.js b/play-return-promise/demo.js
index aa55e7e..2130065 100644
--- a/play-return-promise/demo.js
+++ b/play-return-promise/demo.js
@@ -7,10 +7,12 @@ ChromeSamples.log('Attempting to start playback automatically...');
 startPlayback().then(function() {
   ChromeSamples.log('The play() Promise resolved! Rock on!');
 }).catch(function(error) {
-  ChromeSamples.log('The play() Promise rejected!\nUse the Play button instead.\n' + error);
+  ChromeSamples.log('The play() Promise rejected!\n' +
+                    'Use the Play button instead.\n' + error);

   var playButton = document.querySelector('#play');
-  // The user interaction requirement will be met if playback starts via a click event.
+  // The user interaction requirement will be met if playback starts via a
+  // click event.
   playButton.addEventListener('click', startPlayback);
   playButton.hidden = false;
 });
diff --git a/play-return-promise/index.html b/play-return-promise/index.html
index 86abda4..b0dfd6f 100755
--- a/play-return-promise/index.html
+++ b/play-return-promise/index.html
@@ -17,10 +17,12 @@ feature_id: 5920584248590336
   playback succeeded or failed.
 </p>

-{% capture initial_output_content %}
+{% include output_helper.html %}
+
+{% capture html %}
 <audio id="music" src="../audio/techno.wav"></audio>
 <button id="play" hidden>Play</button>
 {% endcapture %}
-{% include output_helper.html initial_output_content=initial_output_content %}
+{% include html_snippet.html html=html %}

 {% include js_snippet.html filename='demo.js' %}
```

jeffposnick:
Thanks, @beaufortfrancois! I've made changes along the line of what you suggested.

ebidel:
Lgtm

jeffposnick:
Merging. I'll update in a follow-up PR if there is any more feedback that comes out of the article review process.

PR #314, opened 2016-03-11
web-bluetooth: Use gatt.connect instead of connectGATT

tkybpp:
R=@jeffposnick

The warning is shown when use connectGATT, it's deprecated.

Warning message:

> 'BluetoothDevice.connectGATT' is deprecated and will be removed in M52, around August 2016. Please use 'BluetoothDevice.gatt.connect' instead. See https://www.chromestatus.com/feature/5264933985976320 for more details.

### DIFF
--- a/web-bluetooth/battery-level.js
+++ b/web-bluetooth/battery-level.js
@@ -7,7 +7,7 @@ function onButtonClick() {
   .then(device => {
     log('> Found ' + device.name);
     log('Connecting to GATT Server...');
-    return device.connectGATT();
+    return device.gatt.connect();
   })
   .then(server => {
     log('Getting Battery Service...');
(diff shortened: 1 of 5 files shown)


beaufortfrancois:
Pleasw do not merge yet as chrome 50 didn't reach stable channel yet.

jeffposnick:
Is it possible to structure the sample to use feature detection and work in older versions of Chrome as well? I'm not familiar with the API, but something like

``` js
if ('connect' in device.gatt) {
  device.gatt.connect();
} else {
  device.connectGATT();
}
```

beaufortfrancois:
We could do that indeed but this won't be necessary as `device.connectGATT` will work until M52. I'm simply waiting for M50 to reach stable to update all our Web Bluetoooth samples to use `device.gatt.connect`.

tkybpp:
Oh sorry, now I get it.
It occurs error when use `gatt.connect` on Chrome v49 (Chrome OS), right?
But I didn't recognize because I use only Chrome Dev v51.

beaufortfrancois:
No problem @tkybpp.
Chrome M50 is right about the corner now ;)

beaufortfrancois:
Here we are!
Thank you @tkybpp.

tkybpp:
Finally! Thank you!

PR #315, opened 2016-03-14
Add focus navigation start point example

robdodson:


petele:
When this is merged, please add a comment to issue [#2693](https://github.com/google/WebFundamentals/issues/2693)

jeffposnick:
Merging in the interest of it not blocking #2693.

Rob, it would be good to provide in a follow-up PR either a link to the /web/updates/ article or some more context as to what's going on in the page, so that folks who stumble upon the sample directly will have some context. I'll file a follow-up issue assigned to you to make that change.

robdodson:
@jeffposnick yep happy to add more info

Issue #316, opened 2016-03-15
Context for focus navigation start point sample


jeffposnick:
@robdodson, as mentioned in https://github.com/GoogleChrome/samples/pull/315, it would be good to provide some context to what the focus navigation start point sample accomplishes, either inline in the page or via a link to your /web/updates article.

PR #317, opened 2016-03-15
Add more info to the focus navigation example

robdodson:
@jeffposnick PTAL

Fixes https://github.com/GoogleChrome/samples/issues/316


jeffposnick:
Thanks, @robdodson!

PR #319, opened 2016-04-07
Add two-token web session management demo app.

kongguibin:
Add an example Google App Engine app, which leverages service worker to support two-token based web session management.


jeffposnick:
Hello @kongguibin—I'm not sure that this repo is the best home for this kind of sample, given its dependencies on Java App Engine. It's intended purpose is to house samples closely tied to new functionality added to each version of Chrome (tracked at https://www.chromestatus.com/features) that can be demonstrated as part of a static hosted page served from the `gh-pages` branch.

Would you be open to going through the normal Google open source releasing process and creating a new repo for this?

kongguibin:
I changed my primary email address to my google.com one. Hope the CLA is not an issue any more.

kongguibin:
@owencm Owen, can you explain why we select this repository?

PaulKinlan:
I think we should make a new repo, it can be done easily. This is about a
process/method so I think it can standalone.

owencm:
I've started a thread with all of you and Taylor to discuss if this is the right place for this to live. Lets discuss there.

jeffposnick:
Closing, as this has moved to https://github.com/GoogleChrome/two-token-sw

PR #321, opened 2016-04-13
Sample for KeyboardEvent.key

jeffposnick:
R: @PaulKinlan @addyosmani @gauntface etc.

Straightforward example of the new `KeyboardEvent.key` attribute, shipping in M51.

I'm going to put together an updates article covering this and the related `KeyboardEvent.code` feature that shipped in M48.

Live demo available at https://jeffy.info/samples/keyboardevent-key-attribute/


beaufortfrancois:
👍

PR #322, opened 2016-04-14
Check to see if KeyboardEvent.code is supported

jeffposnick:
R: @beaufortfrancois etc.

Just to make it play a little nicer on unsupported browsers.


beaufortfrancois:
:shipit:

PR #324, opened 2016-04-20
Remove the device info and device class fields from BluetoothDevice

beaufortfrancois:
According to https://github.com/WebBluetoothCG/web-bluetooth/pull/229,  device info and device class fields from BluetoothDevice will be removed from now on. Let's remove them from our "Device Info" sample.

screenshot 2016-04-20 at 12 27 29 pm

R=@jeffposnick

### DIFF
--- a/web-bluetooth/device-info.js
+++ b/web-bluetooth/device-info.js
@@ -26,12 +26,8 @@ function onFormSubmit() {
   .then(device => {
     log('> Name:             ' + device.name);
     log('> Id:               ' + device.id);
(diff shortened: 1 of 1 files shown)


jeffposnick:
Quick check: along with the removals, there's an addition of `log('> Connected:        ' + device.gatt.connected);` to the sample. Just wanted to make sure that was intentional. If so, 👍 and feel free to merge.

beaufortfrancois:
It is intentional. Thanks for checking this out though!

PR #331, opened 2016-05-02
s/getElementById/querySelector/g

beaufortfrancois:
For the sake of consistenty among Web Bluetooth samples, I ran `git grep -l "document.getElementById('" | xargs sed -i "s/document.getElementById('/document.querySelector('#/g"`

R=@jeffposnick


jeffposnick:
👍

PR #336, opened 2016-05-11
Service Worker Video Caching Sample

samdutton:
Commits can be squashed on merge :). 

Not sure why the .gitignore file mode has changed :(.


samdutton:
@jeffposnick ptal when you have a moment. Used by https://github.com/google/WebFundamentals/pull/2906.

jeffposnick:
The issue right now with serving videos from the SW cache is that seeking doesn't work out of the box. You need to properly handle `Request`s that include a `Range` header, and construct a specific `Response` using that: https://bugs.chromium.org/p/chromium/issues/detail?id=575357#c10

It would be good to include that technique in this sample, at which point we should be 👍

samdutton:
Hi @jeffposnick — I've added code to handle range requests. 

(Apologies for all the commits — had some trouble at first getting site running locally. Not sure why .gitignore file mode has changed :(.)

jeffposnick:
👍 Thanks, @samdutton!

PR #337, opened 2016-05-13
bluetooth: Add GAP Characteristics Sample

beaufortfrancois:
I'd like to start a new section in our Web Bluetooth Samples for more advanced developers. First one is about accessing and reading GAP Characteristics.

~~Note: This sample works for now with the empty name trick.~~

Preview at https://beaufortfrancois.github.io/samples/web-bluetooth/gap-characteristics.html

R=@g-ortuno @scheib @jyasskin
R=@jeffposnick


beaufortfrancois:
For info, this doesn't work yet on Chrome OS and Chrome for Linux due to a BlueZ issue tracked at https://bugs.chromium.org/p/chromium/issues/detail?id=611678

web-bluetooth/gap-characteristics.js
```diff
@@ -0,0 +1,186 @@
...
+        default: log('> Unknown Characteristic: ' + characteristic.uuid);
+      }
+    });
+  })
+  .catch(error => {
+    log('Argh! ' + error);
+  });
+}
+
+function getDeviceType(value) {
```
jeffposnick, line 105: What do you think about storing this mapping in an 

```
valueToDeviceType = {
  0: 'Unknown',
  ...
};
```

in the outer scope, and then changing this function to

```
return value + (value in valueToDeviceType ?
                ' (' + valueToDeviceType[value] + ')' :
                '');
```

(If you can assume support for template literals, that would be cleaner, too.)
That would lead to a less intimidating `switch()` and you might theoretically be able to reuse that `valueToDeviceType` mapping elsewhere.

web-bluetooth/gap-characteristics.js
```diff
@@ -0,0 +1,186 @@
...
+  .then(service => {
+    log('Getting GAP Characteristics...');
+    return service.getCharacteristics();
+  })
+  .then(characteristics => {
+    characteristics.forEach(characteristic => {
+      switch (characteristic.uuid) {
+
+        case BluetoothUUID.getCharacteristic('gap.appearance'):
+          characteristic.readValue().then(value => {
```
g-ortuno: I think you need to queue these promises otherwise you are going to get "In Progress" errors.

beaufortfrancois:
Thank you @jeffposnick and @g-ortuno.
I believe I addressed your feedback. Please let me know how it looks like now.

web-bluetooth/gap-characteristics.js
```diff
@@ -0,0 +1,107 @@
...
+    log('Getting GAP Characteristics...');
+    return service.getCharacteristics();
+  })
+  .then(characteristics => {
+    let queue = Promise.resolve();
+    characteristics.forEach(characteristic => {
+      switch (characteristic.uuid) {
+
+        case BluetoothUUID.getCharacteristic('gap.appearance'):
+          queue.then(readAppearanceValue(characteristic));
```
g-ortuno: Promises are tricky :sweat_smile: Here, you are calling `readAppearanceValue` immediately rather than waiting for the previous promise to resolve. You need to put the call to `readAppearanceValue` inside another function:

``` js
queue.then(() => readAppearanceValue(characteristic));
```

beaufortfrancois:
@g-ortuno This is now fixed in https://github.com/GoogleChrome/samples/pull/337/commits/ac6cab8f280e5638a29452a470e927f1a90c2898 Thanks!

jeffposnick:
My feedback looks like it's addressed.

web-bluetooth/gap-characteristics.js
```diff
@@ -0,0 +1,107 @@
...
+    log('Getting GAP Characteristics...');
+    return service.getCharacteristics();
+  })
+  .then(characteristics => {
+    let queue = Promise.resolve();
+    characteristics.forEach(characteristic => {
+      switch (characteristic.uuid) {
+
+        case BluetoothUUID.getCharacteristic('gap.appearance'):
+          queue.then(_ => readAppearanceValue(characteristic));
```
g-ortuno: Ah. Sorry for not noticing earlier. You need to do 

``` js
queue = queue.then(...); 
```

Otherwise you are not queuing the promises. I've made all of these errors myself haha.

scheib:
We shouldn't publish official samples/demos that rely on the Chrome bug (name: '') to access all devices. https://github.com/WebBluetoothCG/web-bluetooth/issues/234 should be resolved first, or the code should use a workaround (e.g. on that thread).

beaufortfrancois:
@g-ortuno Fixed in https://github.com/GoogleChrome/samples/pull/337/commits/fecb0e6ba05ab3afd6c7445d8b83cc365de16afe. 

@scheib I agree. Fixed in https://github.com/GoogleChrome/samples/pull/337/commits/836d1b8

g-ortuno:
lgtm. I'm sad that we need to use such nasty hack to get all devices :disappointed:

web-bluetooth/gap-characteristics.js
```diff
@@ -0,0 +1,116 @@
...
+function getDeviceType(value) {
+  // Check out page source to see what valueToDeviceType object is.
+  return value +
+      (value in valueToDeviceType ? ' (' + valueToDeviceType[value] + ')' : '');
+}
+
+function anyDevice() {
+  // This is the closest we can get for now to get all devices.
+  // https://github.com/WebBluetoothCG/web-bluetooth/issues/234
+  return Array.from('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ')
```
scheib, line 114: Also want name: '' to catch an empty named device.

beaufortfrancois, line 114: I thought `name: ''` would not catch anything (eg. bug) since there would never be a empty name as all OS provide a friendly name such as "N/A", "Unknown or unsupported device", etc.

beaufortfrancois:
@scheib Here it is: https://github.com/GoogleChrome/samples/pull/337/commits/7c858340ccdd09a06335883f0ea1d5524b9d86ec

scheib:
LGTM

PR #338, opened 2016-05-17
PaymentRequest sample.

rsolomakhin:
This patch adds a sample for PaymentRequest. Feature page:

https://www.chromestatus.com/features/5639348045217792

Latest spec:

https://github.com/w3c/browser-payment-api


rsolomakhin:
Need to re-upload with rouslan@chromium.org email address. brb.

PR #339, opened 2016-05-17
PaymentRequest samples.

rsolomakhin:
This patch adds several samples for PaymentRequest API. Feature
description:

https://www.chromestatus.com/feature/5639348045217792

Latest spec:

https://github.com/w3c/browser-payment-api


rsolomakhin:
Need to use older eslint syntax for suppression of warnings about PaymentRequest not being defined.

PR #345, opened 2016-05-26
Rename advanced section

beaufortfrancois:
@scheib @jyasskin @g-ortuno Is "Intermediate" better?

### DIFF
--- a/web-bluetooth/index.html
+++ b/web-bluetooth/index.html
@@ -5,7 +5,7 @@ feature_id: 5264933985976320
 icon_url: icon.png
 ---
 
-<p>The following samples show you some of the ways that you can use the <a href="https://github.com/WebBluetoothCG/web-bluetooth">Web Bluetooth API</a></p>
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
gentle ping @jyasskin

addyosmani:
I think intermediate makes sense if you have a good sense for what would now get classified as 'advanced' :) Either way, those cc'd should have a good view on this. LGTM otherwise.

web-bluetooth/index.html
```diff
@@ -17,6 +17,6 @@
 <p><a href="device-disconnect.html">Device Disconnect</a> - disconnect and get notified from a disconnection of a BLE Device after connecting to it.</p>
 <p><a href="get-characteristics.html">Get Characteristics</a> - get all characteristics of an advertised service from a BLE Device.</p>
 
-<h3>Advanced</h3>
+<h3>Intermediate</h3>
```
jyasskin: What if this were titled something about "Combining multiple operations"? From https://github.com/GoogleChrome/samples/pull/360#issuecomment-223963940, it seems like that's the unifying trait that distinguishes samples in this section from the "Beginner" samples.

beaufortfrancois: I like this!

PR #346, opened 2016-05-26
update index.html with Android Pay disclaimer

asieke:
Updated the Android Pay code sample with a disclaimer about Android Pay availability


jeffposnick:
👍

PR #347, opened 2016-05-27
Check for Android and use latest PaymentRequest API

rsolomakhin:
PaymentRequest is implemented only on Android for now. This patch adds a check for Android operating system in the user agent identifier string. If the OS is not Android, the live demo is disabled with a corresponding message.

Also in this patch, the dictionary key `currencyCode` is changed to `currency` in order to match the latest spec.


rsolomakhin:
Need to merge upstream first.

PR #348, opened 2016-05-27
Switch from currencyCode to currency in PaymentRequest

rsolomakhin:
PaymentRequest spec has been updated to use `currency` dictionary key instead of `currencyCode`. The spec change recently landed in Chrome Dev 52, so user are no longer able to run the samples here. This patch updates the samples to use `currency` instead of `currencyCode`. That fixes the samples.


jeffposnick:
👍 Thanks for keeping things up to date.

PR #349, opened 2016-05-30
Add Device Information Advanced Sample

beaufortfrancois:
In the same vein as the GAP Characteristics advanced sample, this one is about the generic Device Information Service which I've found being used in several BLE devices.
This sample features: 
- queuing promises
- get unsigned 16-bit integer
- a blacklisted characteristic
- trick to get all devices

R=@scheib @g-ortuno @jyasskin @jeffposnick


scheib:
LGTM

PR #350, opened 2016-05-30
Add Reconnection Address blacklisted characteristic note

beaufortfrancois:
TBR=@jeffposnick

### DIFF
--- a/web-bluetooth/gap-characteristics.html
+++ b/web-bluetooth/gap-characteristics.html
@@ -13,6 +13,8 @@ with the BLE Peripheral Simulator App from the <a target="_blank"
 href="https://play.google.com/store/apps/details?id=io.github.webbluetoothcg.bletestperipheral">Google
 Play Store</a>.</p>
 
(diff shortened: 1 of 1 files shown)


web-bluetooth/gap-characteristics.html
```diff
@@ -13,6 +13,8 @@
 href="https://play.google.com/store/apps/details?id=io.github.webbluetoothcg.bletestperipheral">Google
 Play Store</a>.</p>
 
+<p>Note: The <a target="_blank" href="https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.gap.reconnection_address.xml">Reconnection Address</a> characteristic is <a target="_blank" href="https://github.com/WebBluetoothCG/registries/blob/7b46fd5d58ea98fa098372a774fd7b981f0bdb0d/gatt_blacklist.txt#L30-L32">blacklisted</a> in Web Bluetooth and forbidden from accessing.</p>
```
jeffposnick, line 16: "forbidden from accessing" is an awkward English construction.

"forbidden from being accessed" would be better, or just leave it out entirely, since you're already saying that it's blacklisted.

PR #351, opened 2016-05-30
bluetooth: Clean parseInt

beaufortfrancois:
Since I only call `parseInt` when it starts with `0x`, I can omit radix as it's implied.

TBR=@jeffposnick


jeffposnick:
👍

PR #353, opened 2016-05-31
Sample for new CSS Flexbox absolute-positioned children.

jpmedley:


jpmedley:
@jeffposnick Unless I'm missing something this is failing on somebody else's files, not mine.

jeffposnick:
The broken build is tracked in https://github.com/GoogleChrome/samples/issues/355 and isn't related to this PR.

css-flexbox-abspos/index.html
```diff
@@ -0,0 +1,32 @@
+---
+feature_name: New CSS Flexbox behavior for absolute-positioned children
+chrome_version: 52
+feature_id: 6600926009753600
+---
+
+<h3>Background</h3>
+<p>A previous version of the CSS Flexible Box Layout specification set the static position of absolute-positioned children as though they were a 0x0 flex item. The latest version of the spec takes them fully out of flow and sets the static position based on align and justify properties.</p>
```
jeffposnick: `<code>align</code>` and `<code>justify</code>` since those are property names.

css-flexbox-abspos/index.html
```diff
@@ -0,0 +1,32 @@
...
+
+<p>The sample CSS and HTML below demonstrate this. Starting with Chrome 52 or later, the green box will be perfectly centered in the red box. In other browsers, the top left corner of the green box will be in the top center of the red box.</p>
```
jeffposnick: Drop the "...or later...", as it doesn't fit with the "Starting with Chrome 52..." sentence opening.

jeffposnick: For the "In other browsers..." bit, is it the case that no other browsers behave the way Chrome 52 does? Maybe just say "In earlier versions of Chrome..." without making a statement about what other browsers have implemented, especially since other browsers might add support in the future.

css-flexbox-abspos/index.html
```diff
@@ -0,0 +1,32 @@
...
+
+{% capture css %}
+/*  */
```
jeffposnick: Remove this empty comment.

css-flexbox-abspos/index.html
```diff
@@ -0,0 +1,32 @@
...
+	justify-content: center;
+}
+.container > * {
+	position: absoulte;
+}
+{% endcapture %}
+{% include css_snippet.html css=css %}
+
+{% capture html %}
+<div class="container" style="background:red; width: 200px; height: 200px;">
```
jeffposnick, line 25: These inline `style` declarations would look better if they were moved to the CSS.

css-flexbox-abspos/index.html
```diff
@@ -0,0 +1,32 @@
...
+
+<h3>Background</h3>
+<p>A previous version of the CSS Flexible Box Layout specification set the static position of absolute-positioned children as though they were a 0x0 flex item. The latest version of the spec takes them fully out of flow and sets the static position based on align and justify properties.</p>
+
+<p>The sample CSS and HTML below demonstrate this. Starting with Chrome 52 or later, the green box will be perfectly centered in the red box. In other browsers, the top left corner of the green box will be in the top center of the red box.</p>
+
+{% capture css %}
+/*  */
+.container {
+	display: flex;
```
jeffposnick: Could you switch to two spaces for indentation?

jeffposnick:
Are you sure that this demo illustrates the change in behavior? I just tried opening it locally using Chrome 51 and again with Chrome 52, and I see the same two boxes in both. Here's the Chrome 51 screenshot:
image

css-flexbox-abspos/index.html
jpmedley: Done.

css-flexbox-abspos/index.html
jpmedley: Done and done.

css-flexbox-abspos/index.html
jpmedley: Thought I was going to use it and forgot.

css-flexbox-abspos/index.html
jpmedley, line 25: Does your mini samples framework have a way to apply style without it being visible in the example? Or, is there any other way to do this? These styles exist to make the result visible, but otherwise have nothing to do with the thing that was change. My intent was to show a CSS example without distractions.

css-flexbox-abspos/index.html
jpmedley: Done.

css-flexbox-abspos/index.html
jeffposnick, line 25: My preference would be to put all the styles together in the inlined, visible CSS, since as a developer I don't expect to see a mix of CSS and inline styles, but if you would prefer, using inline styles is fine.

jpmedley:
I misspelled absolute. That's now fixed. There's one issue I'd like your thoughts on before we proceed.

css-flexbox-abspos/index.html
jeffposnick, line 25: CSS that isn't echoed to the page is possible; just add it to the page outside of a `{% capture ... %}` block. I'd prefer keeping the "hidden" stuff to a minimum, though, so if you don't want to add this to the "visible" CSS, keeping it as an inline style is fine.

jpmedley:
Is there anything else I need to do?

jeffposnick:
👍

Issue #344, opened 2016-05-26
Git fetch/push are sloooooooooow


beaufortfrancois:
Since `service-worker/prefetch-video/static/gbike.webm` and `service-worker/prefetch-video/static/gbike.mp4` large bin files have been added, it is so slow to run git fetch/push...
@jeffposnick Can we do something about this?

jeffposnick:
Adding @samdutton in case he can either cut down on the video quality or length while getting his point across. I'm pretty sure that the need to be hosted locally since the responses are stored in the `Cache Storage API` as same-origin resources.

samdutton:
Tricky. I wanted to use a longer video to show multiple range requests (partials) in action — and I've used pretty much the lowest possible quality and resolution. 

I originally used a smaller video (the one from [simpl.info/v](https://simpl.info/video), 1.9MB), but that gets downloaded in one chunk.

Can we use [Large File Storage](https://git-lfs.github.com/)? Not sure that's relevant (or will work) in this case.

Do the videos definitely need to be stored on the same domain?

If need be, I can go back to the smaller video and just use the WebM version.

jeffposnick:
I'm about 99% sure that what you're doing with the cached `Response` objects (cutting them up when there's a `Range` header) requires either same-origin or CORS to be enabled on a remote origin. Maybe something like Google Cloud Storage could be CORS-enabled, but then you'd have the overhead of maintaining a GCS bucket.

samdutton:
Do we have an existing GCS bucket we could use? [Buckets can be CORS-enabled](https://cloud.google.com/storage/docs/cross-origin#How_CORS_Works).

If not, I'll set one up and try it out.

jeffposnick:
Not as far as I know. (If we did have one, @crhym3 would probably be the one who knew.)

samdutton:
Talked to @crhym3 — will sort on Monday.

beaufortfrancois:
Thank you!

samdutton:
I had some problems with this, but I _think_ that was because the Request mode in the Service Worker was set to 'no-cors'.

Whatever — I changed that to 'cors' (default anyway?) and it seems to work now :). 

[samdutton.github.io/samples/service-worker/prefetch-video](https://samdutton.github.io/samples/service-worker/prefetch-video)

addyosmani:
Can this issue be closed or is there still work being done on it?

samdutton:
Hi @addyosmani — yes, this is fixed and can be closed.

PR #354, opened 2016-06-01
Move prefetch-video videos to Bucket storage

samdutton:
As per request in #344.


service-worker/prefetch-video/service-worker.js
```diff
@@ -68,7 +70,7 @@ self.addEventListener('install', function(event) {
         // (https://slightlyoff.github.io/ServiceWorker/spec/service_worker/index.html#cross-origin-resources)
         // and it is not possible to determine whether an opaque response represents a success or failure
         // (https://github.com/whatwg/fetch/issues/14).
-        var request = new Request(url, {mode: 'no-cors'});
+        var request = new Request(url/* , {mode: 'no-cors'} */);
```
jeffposnick: Instead of commenting this out, can you explicitly delete the code? This is a use case in which CORS is required.

service-worker/prefetch-video/service-worker.js
```diff
@@ -36,9 +36,11 @@ self.addEventListener('install', function(event) {
     '/service-worker/prefetch-video/',
     '/service-worker/prefetch-video/index.js',
     '../../styles/main.css',
-    'static/video.mp4',
-    'static/video.webm',
-    'static/poster.jpg'];
+    'static/poster.jpg',
+    // The videos are stored remotely with CORS enabled.
+    'https://prefetch-video-sample.storage.googleapis.com/gbike.webm',
+    'https://prefetch-video-sample.storage.googleapis.com/gbike.mp4'
```
beaufortfrancois, line 40: May we remove `static/video.mp4` and `static/video.webm` files from the repo as well?

service-worker/prefetch-video/service-worker.js
beaufortfrancois: We should adjust comment above which says "It's very important to use {mode: 'no-cors'}".

beaufortfrancois:
Looks good to me! 💛

jeffposnick:
👍

PR #357, opened 2016-06-01
Fetch referrerPolicy sample

jeffposnick:
R: @addyosmani @beaufortfrancois @gauntface etc.

Augmenting the existing Fetch API samples with a sample for M52's "[Fetch API: referrer policy](https://www.chromestatus.com/feature/5692236742262784)"

### DIFF
--- a/fetch-api/fetch-referrer-policy.html
+++ b/fetch-api/fetch-referrer-policy.html
@@ -1,0 +1,25 @@
+---
+feature_name: Fetch Referrer Policy
+chrome_version: 52
+feature_id: 5692236742262784
+---
+
+<h3>Background</h3>
+<p>
+  This sample shows how requests made via the  Fetch API can use a specific
+  <a href="https://hacks.mozilla.org/2016/03/referrer-and-cache-control-apis-for-fetch/">referrer policy</a>,
+  which affects the value of the <code>Referer</code> HTTP header that's sent
+  to remote servers.
+</p>
+
+{% capture initial_output_content %}
+<p>
+  Multiple <code>fetch()</code> requests are made for a local resource, each with a
+  different <code>referrerPolicy</code> set. You can see the requests and the
+  <code>Referer</code> header associated with each in
+  <a href="https://developer.chrome.com/devtools/docs/network">Chrome DevTools</a>.
+</p>
+{% endcapture %}
+{% include output_helper.html initial_output_content=initial_output_content %}
+
+{% include js_snippet.html filename='fetch-referrer-policy.js' %}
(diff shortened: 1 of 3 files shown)


fetch-api/fetch-referrer-policy.html
```diff
@@ -0,0 +1,25 @@
+---
+feature_name: Fetch Referrer Policy
+chrome_version: 52
+feature_id: 5692236742262784
+---
+
+<h3>Background</h3>
+<p>
+  This sample shows how requests made via the  Fetch API can use a specific
```
beaufortfrancois, line 9: Nit: extra space

fetch-api/fetch-referrer-policy.js
```diff
@@ -0,0 +1,9 @@
+var baseUrl = 'users.json';
+var policies = ['no-referrer', 'no-referrer-when-downgrade', 'origin',
+  'origin-when-cross-origin', 'unsafe-url'];
+
+policies.forEach(policy => {
+  var url = `${baseUrl}?policy=${policy}`;
```
beaufortfrancois: It is confusing as it looks like we can pass `policy` in the URL which is not the case.
How about something like this below?

``` js

var policies = ['no-referrer', 'no-referrer-when-downgrade', 'origin',
    'origin-when-cross-origin', 'unsafe-url'];

policies.forEach(policy => {
  fetch('users.json', {referrerPolicy: policy})
  .then(response => { 
    ChromeSamples.log(`Fetched URL with ${policy}`);
  });
});
```

fetch-api/fetch-referrer-policy.js
```diff
@@ -0,0 +1,9 @@
+var baseUrl = 'users.json';
+var policies = ['no-referrer', 'no-referrer-when-downgrade', 'origin',
```
beaufortfrancois: Maybe a simple comment next to each policy could help.

``` js
var policies = [ 'no-referrer', /* don't reveal who is downloading it */
                 'no-referrer-when-downgrade', ..

```

fetch-api/fetch-referrer-policy.html
jeffposnick, line 9: It's non-printing, intentional two-space indentation, to set the text content apart from the `<p>` parent.

beaufortfrancois, line 9: Noted.

fetch-api/fetch-referrer-policy.js
jeffposnick: Agreed that this could be confusing. Since my guidance is to "look at DevTools" to see what the request looks like, I thought that including the policy in the URL would help identify each specific request.

(It would be easier if we had some way of getting the underlying `Request` object that resulted in a `Response` via `fetch()`, but I don't think that's possible.)

I'll remove the URL parameter.

fetch-api/fetch-referrer-policy.js
jeffposnick: Will add.

jeffposnick:
Merging live now that the review feedback has been addressed.

PR #360, opened 2016-06-03
bluetooth: Add Link Loss Sample

beaufortfrancois:
This sample illustrates the use of the Web Bluetooth API to set the standardized `Alert Level` characteristic from a nearby Bluetooth Device used to expose the current link loss alert level that determines how the device alerts when the link is lost.

Featured: 
- read & write on the same `characteristic`
- `characteristic.value` updated after `characteristic.write` 
- ~~trick to get all devices~~

R=@scheib @jyasskin @g-ortuno @jeffposnick 

screenshot 2016-06-03 at 12 07 35 pm

### DIFF
--- a/web-bluetooth/_includes/value-to-device-type.html
+++ b/web-bluetooth/_includes/value-to-device-type.html
@@ -1,5 +1,5 @@
 <script>
-var valueToDeviceType = {
+const valueToDeviceType = {
      0: 'Unknown',
     64: 'Generic Phone',
    128: 'Generic Computer',
--- a/web-bluetooth/link-loss.html
+++ b/web-bluetooth/link-loss.html
@@ -7,7 +7,7 @@ icon_url: icon.png
 
 {% include_relative _includes/intro.html %}
 
-<p>This sample illustrates the use of the Web Bluetooth API to set the standardized <a
+<p>This sample illustrates the use of the Web Bluetooth API to read and write the standardized <a
 target="_blank"
 href="https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.alert_level.xml">Alert Level</a> characteristic from a nearby Bluetooth
 Device used to expose the current link loss alert level that determines how the
@@ -20,6 +20,7 @@ Play Store</a> and advertise the pre-configured Link Loss service.</p>
 
 <p>
   <select id="alertLevelValue">
+    <option selected disabled hidden>Choose...</option>
     <option value="0x00">0x00 (No Alert)</option>
     <option value="0x01">0x01 (Mild Alert)</option>
     <option value="0x02">0x02 (High Alert)</option>
--- a/web-bluetooth/link-loss.html
+++ b/web-bluetooth/link-loss.html
@@ -11,14 +11,17 @@ icon_url: icon.png
 target="_blank"
 href="https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.alert_level.xml">Alert Level</a> characteristic from a nearby Bluetooth
 Device used to expose the current link loss alert level that determines how the
-device alerts when the link is lost.</p>
+device alerts when the link is lost. You may want to try this demo with the nRF
+Master Control App from the <a target="_blank"
+href="https://play.google.com/store/apps/details?id=no.nordicsemi.android.mcp">Google
+Play Store</a> and advertise the pre-configured Link Loss service.</p>
 
 <button id="readButton">Get Bluetooth Device Alert Level</button>
 
 <p>
   <select id="alertLevelValue">
     <option value="0x00">0x00 (No Alert)</option>
-    <option value="0x01">0x01 (Mid Alert)</option>
+    <option value="0x01">0x01 (Mild Alert)</option>
     <option value="0x02">0x02 (High Alert)</option>
   </select>
   <button id="writeButton" disabled>Set Alert Level</button>
--- a/web-bluetooth/link-loss.html
+++ b/web-bluetooth/link-loss.html
@@ -1,0 +1,43 @@
+---
+feature_name: Web Bluetooth / Link Loss
+chrome_version: 50
+feature_id: 5264933985976320
+icon_url: icon.png
+---
+
+{% include_relative _includes/intro.html %}
+
+<p>This sample illustrates the use of the Web Bluetooth API to set the standardized <a
+target="_blank"
+href="https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.alert_level.xml">Alert Level</a> characteristic from a nearby Bluetooth
+Device used to expose the current link loss alert level that determines how the
+device alerts when the link is lost.</p>
+
+<button id="readButton">Get Bluetooth Device Alert Level</button>
+
+<p>
+  <select id="alertLevelValue">
+    <option value="0x00">0x00 (No Alert)</option>
+    <option value="0x01">0x01 (Mid Alert)</option>
+    <option value="0x02">0x02 (High Alert)</option>
+  </select>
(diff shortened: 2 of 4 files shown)


web-bluetooth/link-loss.js
```diff
@@ -0,0 +1,69 @@
...
+
+function onWriteButtonClick() {
+  if (!alertLevelCharacteristic) {
+    return;
+  }
+  log('Setting Alert Level...');
+  let value = [document.querySelector('#alertLevelValue').value];
+  alertLevelCharacteristic.writeValue(new Uint8Array([value]))
+  .then(_ => {
+    log('> Alert Level: ' +
```
g-ortuno: I would change this to `'Alert Level changed to: '`

web-bluetooth/link-loss.js
```diff
@@ -0,0 +1,69 @@
...
+
+const valueToAlertLevel = {
+    0x00: 'No Alert',
+    0x01: 'Mid Alert',
+    0x02: 'High Alert',
+};
+
+function getAlertLevel(value) {
+  let v = value.getUint8(0);
+  return v + (v in valueToAlertLevel ? ' (' + valueToAlertLevel[v] + ')' : '');
```
g-ortuno: I think `'Unknown'` or `'Invalid'` would be better for non-recognized alert levels.

web-bluetooth/_includes/value-to-device-type.html
```diff
@@ -1,5 +1,5 @@
 <script>
-var valueToDeviceType = {
+const valueToDeviceType = {
```
g-ortuno, line 2: The Google JS style guide says to use all caps e.g. `VALUE_TO_DEVICE_TYPE` for constants. Is that something that should be followed here?

g-ortuno:
lgtm bar a couple of nits

web-bluetooth/link-loss.html
```diff
@@ -0,0 +1,43 @@
...
+href="https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.alert_level.xml">Alert Level</a> characteristic from a nearby Bluetooth
+Device used to expose the current link loss alert level that determines how the
+device alerts when the link is lost.</p>
+
+<button id="readButton">Get Bluetooth Device Alert Level</button>
+
+<p>
+  <select id="alertLevelValue">
+    <option value="0x00">0x00 (No Alert)</option>
+    <option value="0x01">0x01 (Mid Alert)</option>
```
scheib: 'Mild Alert' from https://developer.bluetooth.org/gatt/characteristics/Pages/CharacteristicViewer.aspx?u=org.bluetooth.characteristic.alert_level.xml

scheib:
I'm wondering a bit about 
a) How many devices support this, is this an easy sample to try out with devices folks may have around?
b) What does this sample show that isn't demonstrated already. I think `<option value="0x00">0x00 (No Alert)</option>` is about the only thing unique vs the clearing heart rate energy expended? Call out more clearly in sample list & descriptions what they are particularly good at at showing?

web-bluetooth/link-loss.js
beaufortfrancois: Done.

web-bluetooth/link-loss.js
beaufortfrancois: Indeed. I've added `Unknown`.

web-bluetooth/link-loss.html
beaufortfrancois: Done.
I need 👓

web-bluetooth/_includes/value-to-device-type.html
beaufortfrancois, line 2: It also says `Never use the const keyword as it's not supported in Internet Explorer.`
In this case, the uppercase adds some emphasis on a simple utility object I'd like to keep as minimal as possible, hence the camelCase there.

beaufortfrancois:
> a) How many devices support this, is this an easy sample to try out with devices folks may have around?

I've added a link to the nRF Master Control App which supports the pre-configured Link Loss service. This is obviously not a "real" device that people might have around but it is essentially the same as our own BLE Peripheral Simulator App.

> b) What does this sample show that isn't demonstrated already. I think `<option value="0x00">0x00 (No Alert)</option>` is about the only thing unique vs the clearing heart rate energy expended? Call out more clearly in sample list & descriptions what they are particularly good at at showing?

As mentioned in the description [comment](https://github.com/GoogleChrome/samples/pull/360#issue-158331827), this sample features
- read & write on the same `characteristic`
- `characteristic.value` updated after `characteristic.write` (which is something not known to web developers) 
- trick to get all devices (not unique)

The "beginner" section was about showing how to use one simple method in the Web Bluetooth API. Nothing else. The new section will be about making them work together without necessarily being a Web Bluetooth demo which would require some UI glue we don't want in this repo.

I agree with you though that we should call out more prominently what this sample is about... maybe adding `(readValue/writeValue)` after `Link Loss - set the Alert Level characteristic of a BLE Device.` would be enough. What do you think?

beaufortfrancois:
@g-ortuno @scheib Feedback addressed.

g-ortuno:
lgtm

web-bluetooth/link-loss.js
```diff
@@ -0,0 +1,70 @@
...
+    return service.getCharacteristic('alert_level');
+  })
+  .then(characteristic => {
+    alertLevelCharacteristic = characteristic;
+    document.querySelector('#writeButton').disabled = false;
+    log('Reading Alert Level...');
+    return characteristic.readValue();
+  })
+  .then(value => {
+    log('> Alert Level changed to: ' + getAlertLevel(value));
```
scheib: Consider setting the UI to the initial value read from the remote device? And, this text seems odd, as the level isn't changed - it's just initially found to be this value.

scheib:
Thanks for pointing out again what this sample does in particular. I think it's worth noting the read after write in the page's description and the link to it from the overview page.

web-bluetooth/link-loss.js
beaufortfrancois: 1) I prefer keeping the read value in the `<pre>` log. I've added however a "Choose" default element.
2) My bad, I've changed the wrong string. `'> Alert Level changed to:` is only when writing new value to the Alert Level characteristic. It is fixed now.

beaufortfrancois:
@scheib Last commit addresses some of your feedback. See comments above.

web-bluetooth/link-loss.js
```diff
@@ -0,0 +1,70 @@
+var alertLevelCharacteristic;
+
+function onReadButtonClick() {
+  log('Requesting Bluetooth Device...');
+  navigator.bluetooth.requestDevice(
+    {filters: anyDevice(), optionalServices: ['link_loss']})
```
jyasskin: Why not filter to devices that support the 'link_loss' service? That's a pretty rare service, and I don't see a reason devices would be likely to support it without advertising it.

web-bluetooth/link-loss.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+}
+
+function onWriteButtonClick() {
+  if (!alertLevelCharacteristic) {
+    return;
+  }
+  log('Setting Alert Level...');
+  let value = [document.querySelector('#alertLevelValue').value];
+  alertLevelCharacteristic.writeValue(new Uint8Array([value]))
```
jyasskin, line 39: This builds a doubly-nested array: `new Uint8Array([[thevalue]])` ... which works for a single-value array somehow, but is pretty confusing.

beaufortfrancois, line 39: Fixed.

web-bluetooth/link-loss.js
beaufortfrancois: Fixed.

beaufortfrancois:
I've addressed Jeffrey's feedback. I think we're good to go now!
Thanks everyone for your reviews.

PR #361, opened 2016-06-06
Updates suggested by Jake, plus minor fixes

samdutton:


service-worker/prefetch-video/index.html
```diff
@@ -7,8 +7,9 @@
 
 <video controls  poster="static/poster.jpg"
   style="margin: 0 0 1em 0; max-width: 100%; width: 480px;">
-  <source src="static/gbike.webm" type="video/webm" />
-  <source src="static/gbike.mp4" type="video/mp4" />
+  <!-- Videos are stored remotely with CORS enabled -->
```
beaufortfrancois: We may want to specify that only `https://googlechrome.github.io` has been allowed in CORS Headers to not confuse developers who may try to it this URL and wondering why it doesn't work for them.

beaufortfrancois:
LGTM but one nit.

beaufortfrancois:
@samdutton Is this PR still valid?

samdutton:
Hmm... Not sure what's happened here. 

AFAICT the only code that doesn't seem to be in the actual live demo at https://googlechrome.github.io/samples/service-worker/prefetch-video is the `console.log()` call — but Jeff may not have wanted to use that.

Looks like we should abandon this PR.

@jeffposnick Does that make sense?

jeffposnick:
I'm fine with abandoning if that's your recommendation.

PR #362, opened 2016-06-08
Remove runtime HTTPS redirection

jeffposnick:
R: @addyosmani @gauntface @beaufortfrancois etc.

As per https://github.com/blog/2186-https-for-github-pages we can enforce HTTPS for `gh-pages` on the repo-level (which I've turned on already).


addyosmani:
LGTM and yay!

beaufortfrancois:
Woohoo!

PR #364, opened 2016-06-09
bluetooth: Add get-primary-services sample

beaufortfrancois:
Pending https://codereview.chromium.org/2027973002/

R=@g-ortuno @scheib @jyasskin @jeffposnick 

screenshot 2016-06-09 at 10 42 49 am


web-bluetooth/get-primary-services.js
```diff
@@ -0,0 +1,30 @@
...
+  })
+  .then(server => {
+    // Note that we could also get all services that match a specific UUID by
+    // simply passing it to getPrimaryServices().
+    log('Getting Services...');
+    return server.getPrimaryServices();
+  })
+  .then(services => {
+    log('> Services: ' +
+      services.map(s => s.uuid).join('\n' + ' '.repeat(19)));
```
g-ortuno: why 19?

g-ortuno:
Two thoughts:
1. I wonder if it would be better to have two fields: one dropdown list, like the one we use in other samples, and another text input field for custom UUIDs. Having to write 'heart_rate' every time I want to use the sample is going to be a bit annoying. Or have 'Custom UUID' as one of the options in the list and when that is selected the text input field appears. That requires more UI code, so I would be OK with just two fields.
2. I'm not sure how useful this will be. You basically just get whatever services you put in. A combined sample with getCharacteristics might be better?

web-bluetooth/get-primary-services.js
beaufortfrancois: In that case, it should `12` not `19` spaces. Thanks!
See https://googlechrome.github.io/samples/web-bluetooth/get-characteristics.html for the alignment

screenshot 2016-06-13 at 1 25 18 pm

beaufortfrancois:
> 1. I wonder if it would be better to have two fields: one dropdown list, like the one we use in other samples, and another text input field for custom UUIDs. Having to write 'heart_rate' every time I want to use the sample is going to be a bit annoying. Or have 'Custom UUID' as one of the options in the list and when that is selected the text input field appears. That requires more UI code, so I would be OK with just two fields.

Let's give it a try.

> 1. I'm not sure how useful this will be. You basically just get whatever services you put in. A combined sample with getCharacteristics might be better?

Indeed that was what I was afraid of... I'm afraid the combined sample will look a lot like the Get Characteristics Beginner Sample though. Let's try though.

beaufortfrancois:
@g-ortuno I've taken in account your feedback. WDYT?

g-ortuno:
Can you add another screenshot?

beaufortfrancois:
@g-ortuno, here it is:

screenshot 2016-06-14 at 5 19 18 am

beaufortfrancois:
And you can try it at https://beaufortfrancois.github.io/samples/web-bluetooth/discover-services-and-characteristics.html

web-bluetooth/discover-services-and-characteristics.js
```diff
@@ -0,0 +1,58 @@
...
+      return Promise.reject('Please enter some services first!');
+    }
+    // Note that we could also get all services that match a specific UUID by
+    // simply passing it to getPrimaryServices().
+    log('Getting Services...');
+    return server.getPrimaryServices();
+  })
+  .then(services => {
+    log('Getting Characteristics...');
+    return Promise.all(services.map(service => service.getCharacteristics()))
```
g-ortuno: I was playing around with this code a bit. What do you think?

``` js
services.forEach(service => {
  service.getCharacteristics().then(chars => {
    log('> Service:  ' + service.uuid);
    chars.forEach(char => {
       log('>> Characteristic: ' + characteristic.uuid + ' ' + getSupportedProperties(characteristic));
    }
  }
}
```

web-bluetooth/discover-services-and-characteristics.js
```diff
@@ -0,0 +1,58 @@
...
+        log('> Service: ' + service.uuid);
+        allCharacteristics[i].forEach(characteristic => {
+          let supportedProperties = [];
+          // We only want to print supported properties.
+          for (const p in characteristic.properties) {
+            if (characteristic.properties[p]) {
+              supportedProperties.push(p);
+            }
+          }
+          log('> Characteristic: ' + characteristic.uuid +
```
g-ortuno: nit: Should this be `>> Characteristic:` to indicate they belong to the service?

web-bluetooth/discover-services-and-characteristics.js
beaufortfrancois: Much clearer indeed! Thanks!

web-bluetooth/discover-services-and-characteristics.js
beaufortfrancois: Done.

beaufortfrancois:
@g-ortuno I've include all your feedback and it looks a lot nicer. Thanks! I'll merge it now.
screenshot 2016-06-15 at 10 09 46 am

beaufortfrancois:
It is now publicly accessible at https://googlechrome.github.io/samples/web-bluetooth/discover-services-and-characteristics.html

beaufortfrancois:
Thinking more about it, the refactoring doesn't handle properly the rejected promise of `getCharacteristics()`.... I'm not sure if adding a `catch`  is better than returning `Promise.all`.

beaufortfrancois:
One way to solve this is to adding a `catch`. WDYT @g-ortuno?
I'm not a big fan of it as it now has 2 catches to handle errors when really only one is needed. 

``` diff
diff --git a/web-bluetooth/discover-services-and-characteristics.js b/web-bluetooth/discover-services-and-characteristics.js
index b2b1d14..9659c52 100644
--- a/web-bluetooth/discover-services-and-characteristics.js
+++ b/web-bluetooth/discover-services-and-characteristics.js
@@ -31,6 +31,9 @@ function onButtonClick() {
           log('>> Characteristic: ' + characteristic.uuid + ' ' +
               getSupportedProperties(characteristic));
         });
+      })
+      .catch(error => {
+        log('Argh! ' + error);
       });
     });
   })
```

beaufortfrancois:
Or a Promise `queue` like we did before:

``` diff
diff --git a/web-bluetooth/discover-services-and-characteristics.js b/web-bluetooth/discover-services-and-characteristics.js
index b2b1d14..e522934 100644
--- a/web-bluetooth/discover-services-and-characteristics.js
+++ b/web-bluetooth/discover-services-and-characteristics.js
@@ -24,15 +24,17 @@ function onButtonClick() {
   })
   .then(services => {
     log('Getting Characteristics...');
+    let queue = Promise.resolve();
     services.forEach(service => {
-      service.getCharacteristics().then(characteristics => {
+      queue = queue.then(_ => service.getCharacteristics().then(characteristics => {
         log('> Service: ' + service.uuid);
         characteristics.forEach(characteristic => {
           log('>> Characteristic: ' + characteristic.uuid + ' ' +
               getSupportedProperties(characteristic));
         });
-      });
+      }));
     });
+    return queue;
   })
   .catch(error => {
     log('Argh! ' + error);
```

And I think I like this one better. So I'll go with it for now. Please let me know what you think.

mmagalhaes93:
Hi @beaufortfrancois ! i'm gettinng some trouble in getPrimaryService function. It is not retrieving anything and i don't understand why. Could you help me? Thank in advance
screenshot from 2017-10-24 16-14-58

PR #365, opened 2016-06-10
Update PaymentRequest API for Chrome Dev 53.0.2763.0

rsolomakhin:
The 'items' field has been split into 'total' and 'displayItems'. Both
of these fields no longer require 'id'.

The 'shippingOptions' field now lets developers explicitly control which
shipping option is currently selected via the 'selected' boolean field.
'shippingOptions' still requires 'id', so the browser can tell the
website which shipping option the user ended up selecting.

The 'shippingAddress' is now also returned inside of the
'instrumentResponse'. Some of its fields have been renamed to be more
developer friendly and new fields have been added.
- 'region' -> 'country'
- 'administrativeArea' -> 'region'
- 'locality' -> 'city'
- New field 'careOf'
- New field 'phone'

This patch also includes an update to feature detection with more
detailed status message and a note that these tests are for Chrome 53.
Chrome 52 has an older version of PaymentRequest, but there's no sense
in keeping multiple version of examples here.

### DIFF
--- a/paymentrequest/android-pay/demo.js
+++ b/paymentrequest/android-pay/demo.js
@@ -2,21 +2,15 @@ function onBuyClicked() {
   var supportedInstruments = ['https://android.com/pay'];
 
   var details = {
-    items: [
+    total: {label: 'Donation', amount: {currency: 'USD', value: '55.00'}},
+    displayItems: [
       {
-        id: 'original',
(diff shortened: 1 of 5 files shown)


rsolomakhin:
Need to ignore the negated condition warning...

PR #367, opened 2016-06-13
Add Presentation API samples.

lauweijie:
Adds a simple demo for the Presentation API.

Feature page: https://www.chromestatus.com/feature/6676265876586496
Latest spec: http://w3c.github.io/presentation-api/

cc: @japacible @mfoltzgoogle

### DIFF
--- a/presentation-api/index.html
+++ b/presentation-api/index.html
@@ -52,18 +52,6 @@ function updatePresentationAvailability(displaysAvailable) {
       (displaysAvailable ? 'available' : 'unavailable') + '.');
 }
 
-function setConnection(connection) {
-  presentationConnection = connection;
-  document.querySelector('#close').disabled = false;
-  document.querySelector('#terminate').disabled = false;
-  presentationConnection.onclose = function() {
-    ChromeSamples.log('Connection closed.');
-  };
-  presentationConnection.onterminate = function() {
-    ChromeSamples.log('Connection terminated.');
-  };
-}
-
 document.querySelector('#createRequest').addEventListener('click', function() {
   presentationRequest = new PresentationRequest(presentationUrl);
   document.querySelector('#createRequest').disabled = true;
@@ -77,6 +65,18 @@ document.querySelector('#createRequest').addEventListener('click', function() {
   });
 });
 
+function setConnection(connection) {
+  presentationConnection = connection;
+  document.querySelector('#close').disabled = false;
+  document.querySelector('#terminate').disabled = false;
+  presentationConnection.onclose = function() {
+    ChromeSamples.log('Connection closed.');
+  };
+  presentationConnection.onterminate = function() {
+    ChromeSamples.log('Connection terminated.');
+  };
+}
+
 document.querySelector('#start').addEventListener('click', function() {
   presentationRequest.start().then(function(connection) {
     ChromeSamples.log('Connected to presentation: ' + connection.id);
@@ -88,11 +88,11 @@ document.querySelector('#start').addEventListener('click', function() {
 
 document.querySelector('#reconnect').addEventListener('click', function() {
   var presentationId = document.querySelector('#presentationId').value.trim();
-  if (presentationId != '') {
+  if (presentationId.length) {
     presentationRequest.reconnect(presentationId).then(function(connection) {
       ChromeSamples.log('Reconnected to presentation: ' + connection.id);
       setConnection(connection);
-    }).catch(function(error){
+    }).catch(function(error) {
       ChromeSamples.log(error.name + ': ' + error.message);
     });
   }
--- a/presentation-api/index.html
+++ b/presentation-api/index.html
@@ -19,48 +19,52 @@ feature_id: 6676265876586496
 
 {% capture initial_output_content %}
 <p>
-  <button id="newPresReq">new PresentationRequest()</button>
+  <button id="createRequest">new PresentationRequest()</button>
 </p>
 <p>
-  <button id="startPres" disabled>presentationRequest.start()</button>
+  <button id="start" disabled>presentationRequest.start()</button>
 </p>
 <p>
   <input id="presentationId" type="text" placeholder="presentationId">
-  <button id="reconnectPres" disabled>presentationRequest.reconnect(presentationId)</button>
+  <button id="reconnect" disabled>presentationRequest.reconnect(presentationId)</button>
 </p>
 <p>
-  <button id="closePres" disabled>presentationRequest.close()</button>
-  <button id="terminatePres" disabled>presentationRequest.terminate()</button>
-</p>{% endcapture %}
+  <button id="close" disabled>presentationRequest.close()</button>
+  <button id="terminate" disabled>presentationRequest.terminate()</button>
+</p>
+{% endcapture %}
 {% include output_helper.html initial_output_content=initial_output_content %}
 
 {% capture js %}
+// App ID EF1A139F is registered in the Google Cast SDK Developer
+// Console and points to the following custom receiver:
+// https://googlechrome.github.io/samples/presentation-api/receiver/index.html
 var presentationUrl = 'http://google.com/cast/#__castAppId__=EF1A139F';
 var presentationRequest;
 var presentationConnection;
 
 function updatePresentationAvailability(displaysAvailable) {
   ChromeSamples.log('Presentation displays are ' +
-          (displaysAvailable ? 'available' : 'unavailable') + '.');
+      (displaysAvailable ? 'available' : 'unavailable') + '.');
 }
 
 function setConnection(connection) {
-    presentationConnection = connection;
-    document.querySelector('#closePres').disabled = false;
-    document.querySelector('#terminatePres').disabled = false;
-    presentationConnection.onclose = function() {
-      ChromeSamples.log('Connection closed.');
-    };
-    presentationConnection.onterminate = function() {
-      ChromeSamples.log('Connection terminated.');
-    };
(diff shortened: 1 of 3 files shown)


presentation-api/index.html
```diff
@@ -0,0 +1,111 @@
...
+  which gives the ability to access external presentation-type displays and use
+  them for presenting web content. The <code>PresentationRequest</code> object
+  is associated with a request to initiate or reconnect to a presentation made
+  by a controlling browsing context and it takes in a presentation request URL
+  when constructed. A presentation can be started by calling the
+  <code>start()</code> method on the <code>PresentationRequest</code> object.
+</p>
+
+{% capture initial_output_content %}
+<p>
```
japacible, line 25: Use `<div>` instead of `<p>` since this isn't a paragraph.

presentation-api/index.html
```diff
@@ -0,0 +1,111 @@
...
+  <button id="newPresReq">new PresentationRequest()</button>
```
japacible: nit: here and below, "Pres" can be removed from the ids. e.g. "createNewRequest", "start", "reconnect".

presentation-api/index.html
```diff
@@ -0,0 +1,111 @@
...
+  <button id="reconnectPres" disabled>presentationRequest.reconnect(presentationId)</button>
+</p>
+<p>
+  <button id="closePres" disabled>presentationRequest.close()</button>
+  <button id="terminatePres" disabled>presentationRequest.terminate()</button>
+</p>{% endcapture %}
+{% include output_helper.html initial_output_content=initial_output_content %}
+
+{% capture js %}
+var presentationUrl = 'http://google.com/cast/#__castAppId__=EF1A139F';
```
japacible, line 46: Add a comment explaining where the app ID comes from.

presentation-api/index.html
```diff
@@ -0,0 +1,111 @@
...
+var presentationRequest;
+var presentationConnection;
+
+function updatePresentationAvailability(displaysAvailable) {
+  ChromeSamples.log('Presentation displays are ' +
```
japacible, line 51: nit: here and below, fix indentation.

japacible:
Does this work without a polyfill? If not, make a note of that and/or include a polyfill.

presentation-api/index.html
lauweijie, line 25: I used a `<p>` tag so that there will be some margin between the form fields without the need for extra css markup. 

00pqy6zpp8b
veamyeafndh

w3c also [considers each part of a form a paragraph](https://www.w3.org/TR/html5/forms.html#writing-a-form's-user-interface):

> Each part of a form is considered a paragraph, and is typically separated from other parts using p elements.

presentation-api/index.html
lauweijie, line 51: Fixed, thanks.

lauweijie:
@japacible The demo works without a polyfill when Media Router is enabled. I added a note to point this out to users.

japacible:
LGTM @PaulKinlan

beaufortfrancois:
I'd like to give a try to this sample before reviewing it and as I'm debugging on `localhost:4000`, it looks like `PresentationAvailability` value is `false`. Is it because App ID EF1A139F can only work with  https://googlechrome.github.io/samples/presentation-api/receiver/index.html?

presentation-api/index.html
```diff
@@ -0,0 +1,118 @@
...
+    ChromeSamples.log('Connection terminated.');
+  };
+}
+
+document.querySelector('#createRequest').addEventListener('click', function() {
+  presentationRequest = new PresentationRequest(presentationUrl);
+  document.querySelector('#createRequest').disabled = true;
+  document.querySelector('#start').disabled = false;
+  document.querySelector('#reconnect').disabled = false;
+  presentationRequest.getAvailability().then(function(availability) {
```
beaufortfrancois, line 60: Can `getAvailability` reject? If so we might want, a`catch` there as well.

presentation-api/index.html
```diff
@@ -0,0 +1,118 @@
...
+  document.querySelector('#terminate').disabled = false;
+  presentationConnection.onclose = function() {
+    ChromeSamples.log('Connection closed.');
+  };
+  presentationConnection.onterminate = function() {
+    ChromeSamples.log('Connection terminated.');
+  };
+}
+
+document.querySelector('#createRequest').addEventListener('click', function() {
```
beaufortfrancois, line 55: You may want to move this function above `setConnection` as `document.querySelector('#createRequest')` is the first button user is supposed to click. It will help to read/navigate through all functions.

presentation-api/index.html
```diff
@@ -0,0 +1,118 @@
...
+    ChromeSamples.log('Connected to presentation: ' + connection.id);
+    setConnection(connection);
+  }).catch(function(error) {
+    ChromeSamples.log(error.name + ': ' + error.message);
+  });
+});
+
+document.querySelector('#reconnect').addEventListener('click', function() {
+  var presentationId = document.querySelector('#presentationId').value.trim();
+  if (presentationId != '') {
```
beaufortfrancois: nit: `if (presentationId.length) {`

presentation-api/index.html
```diff
@@ -0,0 +1,118 @@
...
+    presentationRequest.reconnect(presentationId).then(function(connection) {
+      ChromeSamples.log('Reconnected to presentation: ' + connection.id);
+      setConnection(connection);
+    }).catch(function(error){
```
beaufortfrancois: nit: `}).catch(function(error) {`

presentation-api/index.html
```diff
@@ -0,0 +1,118 @@
...
+    ChromeSamples.log('Connection closed.');
+  };
+  presentationConnection.onterminate = function() {
+    ChromeSamples.log('Connection terminated.');
+  };
+}
+
+document.querySelector('#createRequest').addEventListener('click', function() {
+  presentationRequest = new PresentationRequest(presentationUrl);
+  document.querySelector('#createRequest').disabled = true;
```
beaufortfrancois, line 57: I would run this enable/disable buttons functions only if `availability.value` is `true`.

beaufortfrancois:
Is it possible to also add another sample that shows how to use the presentation API without Google Cast?

presentation-api/index.html
lauweijie, line 60: According to the specs, the only case where `getAvailability` rejects is if the user agent is unable to continuously monitor the list of available presentation displays but can find presentation displays in order to start a connection. When that happens, the promise is rejected with a `NotSupportedError` exception. I could add a `catch` block with a comment, but the promise would probably never reject for Chrome.

presentation-api/index.html
lauweijie, line 57: I could, but that way I will not be able to demonstrate what happens if `start()` is called when no displays are available -- which I thought could be more helpful than merely disabling the button. What are your thoughts?

bgqhntl f8cv j0kbtu aaaaaelftksuqmcc

lauweijie:
@beaufortfrancois Are you trying the demo with Media Router enabled? The demo does not rely on the  Cast SDK and should not require the Google Cast extension to be installed. If MR is not enabled (which seems to be the case here), then `PresentationAvailability` would be `false`.

The correct behavior should be the GitHub 404 page being cast to the device (until the PR is merged in).

`chrome://flags/#media-router`

beaufortfrancois:
I am trying the demo with `--media-router=1` enabled at `chrome://flags/#media-router` on `localhost`.  Here's what happens when I click on `new PresentationRequest()` button:

Note that clicking on `start()` button displays the Media Router UI though.

screenshot 2016-06-17 at 4 23 20 pm

presentation-api/index.html
beaufortfrancois, line 60: Makes sense. Noted.

presentation-api/index.html
beaufortfrancois, line 57: Oh you're right!

lauweijie:
> Note that clicking on `start()` button displays the Media Router UI though.

@beaufortfrancois Does your device show up in the Media Router UI? If it does, the `presentationAvailability` value should be `true`. I can't seem to reproduce a case where the device shows up in the MR UI and `presentationAvailability.value` is `false`.

Update: There is one case where that happens. If there are no receivers that supports the castAppId provided (happens for apps which are not published, devices that are not whitelisted or invalid castAppIds), the `presentationAvailability` value is false but the MR dialog shows up with devices in the list. Clicking on the device mirrors the page instead of casting the URL. Is that what is happening in your case?

beaufortfrancois:
> @beaufortfrancois Does your device show up in the Media Router UI? If it does, the presentationAvailability value should be true. I can't seem to reproduce a case where the device shows up in the MR UI and presentationAvailability.value is false.

No devices show up as none is plugged intentionally.
That's why I'm confused between `new PresentationRequest()` and `presentationRequest.start()` which works without picked devices...

lauweijie:
According to the draft specs, `presentationRequest.start()` should by right reject the promise with a NotFoundError DOMException if no displays are available. I had a quick chat with someone on my team and [filed a bug](https://bugs.chromium.org/p/chromium/issues/detail?id=621161) for it in Chome. The specs for the Presentation API is still evolving and its implementation is partial. As far as I know, the cause of it should not be with this particular demo. Thanks for looking through and testing the demo.

beaufortfrancois:
LGTM 👍

beaufortfrancois:
And it's now live at https://googlechrome.github.io/samples/presentation-api/

PR #369, opened 2016-06-15
Add warning message if min chrome version check fails

beaufortfrancois:
Hello Jeff,
I believe the simple warning added in this CL will help developers who want to try out our samples but are not on Canary for instance.

screenshot 2016-06-15 at 11 18 13 am

R=@jeffposnick

### DIFF
--- a/_includes/output_helper.html
+++ b/_includes/output_helper.html
@@ -33,12 +33,14 @@
   <pre id="log"></pre>
 </div>
 
+{% if page.check_min_version %}
 <script>
   if (/Chrome\/(\d+\.\d+.\d+.\d+)/.test(navigator.userAgent)){
     // Let's log a warning if the sample is not supposed to execute on this
     // version of Chrome.
     if ({{ page.chrome_version }} > parseInt(RegExp.$1)) {
-      ChromeSamples.setStatus('Warning! This sample requires Chrome ' + {{ page.chrome_version }} + '.');
+      ChromeSamples.setStatus('Warning! Keep in mind this sample has been tested with Chrome ' + {{ page.chrome_version }} + '.');
     }
   }
 </script>
+{% endif %}
--- a/_includes/output_helper.html
+++ b/_includes/output_helper.html
@@ -32,3 +32,13 @@
   <div id="status"></div>
   <pre id="log"></pre>
 </div>
+
+<script>
+  if (/Chrome\/(\d+\.\d+.\d+.\d+)/.test(navigator.userAgent)){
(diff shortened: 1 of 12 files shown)


jeffposnick:
My preference would be not to go this route. While I understand that this code will only run if it thinks the user agent is already (and older) Chrome, I don't like the message that the sample *requires" Chrome, since ideally we're only writing samples for standard features that will eventually make it to all browsers.

I also am worried that there are samples out there that are targeted to version X, but have either a polyfill or a workaround in place to show how they could be used in version < X, and we'd end up scaring developers away with that status.

beaufortfrancois:
> My preference would be not to go this route. While I understand that this code will only run if it thinks the user agent is already (and older) Chrome, I don't like the message that the sample *requires" Chrome, since ideally we're only writing samples for standard features that will eventually make it to all browsers.

How about rewording the message to something like `Warning! Keep in mind this sample has been tested with Chrome 54.` ?

> I also am worried that there are samples out there that are targeted to version X, but have either a polyfill or a workaround in place to show how they could be used in version < X, and we'd end up scaring developers away with that status.

jeffposnick:
I don't think that would address my second concern, which is that some samples are written to explicitly illustrate backwards compatibility. For example, https://github.com/GoogleChrome/samples/blob/gh-pages/idb-getall/demo.js#L46

If you'd like to add this into a template that's specific to the Bluetooth samples, where you know the backwards compatibility story is unlikely to change, then I'm cool with that.

beaufortfrancois:
I see... ;(
I think I'll add a boolean setting that each sample can define to display a warning if min chrome version is not met yet.

beaufortfrancois:
Hey @jeffposnick What do you think of https://github.com/GoogleChrome/samples/pull/369/commits/bfc3dcde62235fd5362be6e95f7a1be067960c93?

jeffposnick:
👍

Psoffie009:
> Hello Jeff,
> I believe the simple warning added in this CL will help developers who want to try out our samples but are not on Canary for instance.
> 
> screenshot 2016-06-15 at 11 18 13 am
> 
> R=@jeffposnick

Psoffie009:
> My preference would be not to go this route. While I understand that this code will only run if it thinks the user agent is already (and older) Chrome, I don't like the message that the sample *requires" Chrome, since ideally we're only writing samples for standard features that will eventually make it to all browsers.
> 
> I also am worried that there are samples out there that are targeted to version X, but have either a polyfill or a workaround in place to show how they could be used in version < X, and we'd end up scaring developers away with that status.

PR #370, opened 2016-06-15
Update README.md

ghost:
Make it explicitly clear that the service worker script should reside in the root directory, I've put mine in /scripts/sw.js and was stuck in cached files not being updated left scratching my head why is this happening.

### DIFF
--- a/service-worker/README.md
+++ b/service-worker/README.md
@@ -1,3 +1,11 @@
+# Before you begin
+
+>For security reasons, a service worker can only control the pages that are in the same directory level or below it. This means that if you place the service worker file in a scripts directory it will be only able to be attached to pages inside the /scripts directory and below (/scripts/test/ for example). It is unlikely that your pages for your site will live there.
+
+[By](https://developers.google.com/web/fundamentals/getting-started/your-first-offline-web-app/step-5?hl=en) @paulirish
+
+Make sure you've read the above quote.
+
 # Service Worker Recipes
 
 - [Basic registration](https://googlechrome.github.io/samples/service-worker/registration/index.html) -


ghost:
@googlebot Calculate -0 + -'aha

service-worker/README.md
```diff
@@ -1,3 +1,11 @@
+# Before you begin
+
+>For security reasons, a service worker can only control the pages that are in the same directory level or below it. This means that if you place the service worker file in a scripts directory it will be only able to be attached to pages inside the /scripts directory and below (/scripts/test/ for example). It is unlikely that your pages for your site will live there.
+
+[By](https://developers.google.com/web/fundamentals/getting-started/your-first-offline-web-app/step-5?hl=en) @paulirish
```
jeffposnick, line 5: That's @PaulKinlan.

jeffposnick:
I'm sorry that this restriction caused confusion when you were experimenting in your own site.

My thinking is that there's a lot of relevant background information about service workers. Rather than explicitly calling out a subset of it here, how about linking to, for example, http://www.html5rocks.com/en/tutorials/service-worker/introduction/#toc-before and telling folks to read that first?

jpmedley:
Technical writer here. Agree strongly, Jeff.

ghost:
I'm going through some real-life difficulties right now. Please hijack this issue and deal with it.

beaufortfrancois:
@jeffposnick Is this still relevant? If not, let's close this issue.

jeffposnick:
I believe addressing this as part of the overall revamp of service worker material on https://developers.google.com/web/ would make the most sense. (CC: @jpmedley)

PR #371, opened 2016-06-16
bluetooth: Add automatic reconnect sample

beaufortfrancois:
This sample provides some guidance on how to automatically reconnect to a disconnected device using a simple exponential backoff algorithm. We can also use it as a way to calculate how much time scan will timeout. For instance, the reconnect algorithm is useless for now on macOS as `device.gatt.connect` never (ever) rejects.

You can try it live at https://beaufortfrancois.github.io/samples/web-bluetooth/automatic-reconnect.html

R=@g-ortuno @scheib @jyasskin

screenshot 2016-06-16 at 3 14 06 pm

### DIFF
--- a/web-bluetooth/automatic-reconnect.html
+++ b/web-bluetooth/automatic-reconnect.html
@@ -1,0 +1,35 @@
+---
+feature_name: Web Bluetooth / Automatic Reconnect
+chrome_version: 50
+feature_id: 5264933985976320
+icon_url: icon.png
+---
+
+{% include_relative _includes/intro.html %}
+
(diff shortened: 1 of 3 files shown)


scheib:
LGTM

beaufortfrancois:
I'll wait for https://bugs.chromium.org/p/chromium/issues/detail?id=621022 to be fixed before committing this sample.

beaufortfrancois:
https://bugs.chromium.org/p/chromium/issues/detail?id=621022#c8 is now fixed so let's push it!

PR #372, opened 2016-06-16
CSS Nit: Align vertically status and log

beaufortfrancois:
Hey @jeffposnick,
This is just a nit that will improve Live Output.


addyosmani:
LGTM but deferring to Jeff

jeffposnick:
👍

Issue #376, opened 2016-06-24
Add ability to cache samples


beaufortfrancois:
Hey @jeffposnick,

What do you think of adding the ability for each sample to be properly cached offline?
For the Web Bluetooth samples for instance, it would make a lot of sense.

I like the simple strategy stale-while-revalidate (https://paul.kinlan.me/my-blogs-service-worker-and-caching-strategy/). Any thoughts?

addyosmani:
@beaufortfrancois What's the use-case you had in mind for (non-SW) samples supporting full offline caching? Outside of PWAs, I can see value in codelab samples/walkthroughs working offline but naively assume folks looking at samples from this repo just use it for reference / don't run on device that much.

beaufortfrancois:
Web Bluetooth samples are maybe an exception as we use them a lot to test out BLE devices and offline can come handy. Check out our collection at https://googlechrome.github.io/samples/web-bluetooth/.

jeffposnick:
I'm not a huge fan of stale-while-revalidate for this sort of templated site.

There is some cache duplication due to partial resources that would all be repeated in the composed HTML page, though those are admittedly very small and wouldn't waste a meaningful amount of storage. More of an issue is that the stale-while-revalidation would apply to each individual page instead of applying once to the shared partial templates. So if a user visits page A, then B, then we update the partial template, the next time they visit A they'll still see the old partial template, and then the next time they visit B they'll still see the old partial template, since the revalidation only applied to the composed HTML in page A.

I am in favor of putting together a service worker for the site that cached our full and partial templates, and did the composition inside the service worker. A quick search turned up a couple of JavaScript implementations of the Liquid templating language that `gh-pages` uses via the Jekyll build process:
- https://github.com/heupel/liquid.js
- https://github.com/mattmccray/liquid.js

(They seem to render to strings rather than progressively to a `ReadableStream`, but that's not a hard requirement.)

My objections to stale-while-revalidate aren't very strenuous. It's mainly that service workers for this sort of templated site is very much on everyone's mind right now, and I'd vote for doing what we'd end up recommending as a best-practice if we're going to do anything at all.

beaufortfrancois:
As you've noted, the stale-while-revalidate pattern was just a first step to get a nice service-worker for our Chrome samples.The only thing I would be afraid of with a service worker doing the composition is that it would differ with the actual generated html file from Jekyll.

jeffposnick:
[...]

PR #380, opened 2016-06-27
Add Web Bluetooth service worker

beaufortfrancois:
@jeffposnick @addyosmani  In order to provide offline access to Web Bluetooth samples, I've added a simple service worker (stale while revalidate pattern). If that looks good to you, we could extend this solution to other samples. See https://github.com/GoogleChrome/samples/issues/376

Live preview at https://beaufortfrancois.github.io/samples/web-bluetooth/index.html

### DIFF
--- a/web-bluetooth/service-worker.js
+++ b/web-bluetooth/service-worker.js
@@ -1,0 +1,26 @@
+self.addEventListener('fetch', function(event) {
+  const request = event.request;
+  const url = new URL(event.request.url);
+
+  // Don't cache anything that is not on this origin.
+  if (url.origin !== location.origin) {
+    return;
+  }
+
+  event.respondWith(
+    caches.open('web-bluetooth').then(cache => {
+      return cache.match(request).then(response => {
+        var fetchPromise = fetch(request).then(networkResponse => {
+          cache.put(request, networkResponse.clone());
+          return networkResponse;
+        });
+        // We need to ensure that the event doesn't complete until we
+        // know we have fetched the data
+        event.waitUntil(fetchPromise);
+
+        // Return the response from cache or wait for network.
+        return response || fetchPromise;
+      });
+    })
+  );
+});
--- a/web-bluetooth/service-worker.js
+++ b/web-bluetooth/service-worker.js
@@ -7,20 +7,14 @@ self.addEventListener('fetch', function(event) {
     return;
   }
 
-  event.respondWith(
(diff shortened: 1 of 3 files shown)


jeffposnick:
Hey @beaufortfrancois: see my comment in #376.

beaufortfrancois:
Should we proceed then for now with this not-perfect solution?

web-bluetooth/service-worker.js
```diff
@@ -0,0 +1,26 @@
...
+  event.respondWith(
+    caches.open('web-bluetooth').then(cache => {
+      return cache.match(request).then(response => {
+        var fetchPromise = fetch(request).then(networkResponse => {
+          cache.put(request, networkResponse.clone());
+          return networkResponse;
+        });
+        // We need to ensure that the event doesn't complete until we
+        // know we have fetched the data
+        event.waitUntil(fetchPromise);
```
jeffposnick: FYI, this doesn't accomplish anything (in Chrome): https://bugs.chromium.org/p/chromium/issues/detail?id=601095

web-bluetooth/service-worker.js
```diff
@@ -0,0 +1,26 @@
+self.addEventListener('fetch', function(event) {
+  const request = event.request;
+  const url = new URL(event.request.url);
+
+  // Don't cache anything that is not on this origin.
```
jeffposnick, line 5: Why not? What cross-origin resources do you anticipate being fetched?

web-bluetooth/service-worker.js
beaufortfrancois: I've been following @paulkinlan SW strategy there: https://paul.kinlan.me/my-blogs-service-worker-and-caching-strategy/

web-bluetooth/service-worker.js
beaufortfrancois, line 5: Data: URLs such as Google analytics

web-bluetooth/service-worker.js
jeffposnick: You can leave that call in if you'd like, but I'd recommend changing the comment to point readers to that crbug and make it clear that it's not doing anything now.

beaufortfrancois: I've removed it to avoid confusion.

beaufortfrancois:
As discussed offline, I'll merge this PR and we'll figure out how to scale for all chrome samples.

PR #387, opened 2016-07-13
Update Travis CI config to use Ruby 2.2.2

jeffposnick:
R: @beaufortfrancois @addyosmani @PaulKinlan etc.

As per https://travis-ci.org/GoogleChrome/samples/builds/144504931#L193, our Travis CI builds are failing because one of the depedencies expects Ruby 2.2.2. I don't know of any reason not to upgrade, so this change does that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-- 2.1
+- 2.2
 script:
   - npm install
   - npm run lint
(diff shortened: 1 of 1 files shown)


addyosmani:
waiting for the build to complete (that last commit is still building) but otherwise LGTM.

jeffposnick:
Travis has to download the Ruby 2.2.2 package and install it as part of the build, since it isn't in the Travis environment (yet). But it does seem to work now.

PR #393, opened 2016-07-29
Force Flattening When Ancestor Has Opacity Sample

jeffposnick:
R: @beaufortfrancois @PaulKinlan @surma @ebidel etc.

This is a sample to accompany the "[Force flattening for elements with opacity < 1](https://www.chromestatus.com/feature/5671480339726336)" feature launching in Chrome 53.

Much of the background and the sample is transferred from the [Intent to Ship](https://docs.google.com/document/d/1xdOyUKXYCKKcMJd8fFyuG0JLTi72ajqQx8h1w1MXKWw/edit#) announcement. I can't find the relevant engineers on GitHub to CC: them, so I'll email them a link to this PR as a heads-up.

Live version is staged at https://jeffy.info/samples/css-opacity-force-flattening/index.html


beaufortfrancois:
I believe a with and without `preserve-3d` view would help to understand there as it confused me. WDYT?

screenshot 2016-08-01 at 10 20 57 am

surma:
I tend to agree, although I think the screenshots should say “before fix” and “after fix” or something. The point is that `opacity` on the container would cause Chrome to render each child element independently with the same opacity setting before the fix.

beaufortfrancois:
@surma 👍

surma:
Left: M51
Right: M53 (also in M52?)
screenshot-2016-08-01-10 10 01_

beaufortfrancois:
I believe the without `preserve-3d` example would still help to understand what it does also.

surma:
I agree, but the question is do people care about this change who do not understand what `preserve-3d` does? I guess it’s a question of target audience. We could just include both :D

ebidel:
I think it would be good to include an image of the expected output of both browsers in the demo. Anytime we're talking render, it's good to have a reference.

I was also going to suggest a supplemental demo like http://desandro.github.io/3dtransforms/examples/cube-02-show-sides.html for this sample.  I'm not sure the current demo really highlights what's going (aka why one would care about this change). Popping `opacity: 0.8` on the container cube in the desandro demo makes things really obvious.

jeffposnick:
I'm going to keep it lightweight and just go with the images of the before/after, along with the live output. I'm having some trouble updating my staging copy, but here's a screenshot of part of the page. (If we think we need to provide more guidance than this, than maybe an article is in order, afterall.)

image

jpmedley:
That reads at first glance like it's changing the z order. I know it
doesn't, but I'm wondering if that needs to be called out in the text.

Joe Medley | Technical Writer, Chrome DevRel | jmedley@google.com |
 816-678-7195
_If an API's not documented it doesn't exist._

jeffposnick:
I don't know whether there are any subtleties when it comes to elements that also happen to have `z-index` set, so I'd rather not make a statement about it one way or another in this sample. (This again falls into the category of things that might make sense to explore if we were writing a full-fledged post about the change.)

jeffposnick:
I'm going to get this merged given that the initial feedback was addressed. If there's any ongoing concerns, I'll address them in a follow-up PR.

PR #394, opened 2016-07-29
Rasterization & will-change: transform sample

jeffposnick:
R: @surma @PaulKinlan @beaufortfrancois @ebidel etc.

This is a sample to accompany the "[Raster on composited layer scale change, except if will-change: transform or an accelerated animation is present](https://www.chromestatus.com/feature/5637351992721408)" feature launching in Chrome 53.

Much of the background and the sample is transferred from the [Intent to Ship](https://docs.google.com/document/d/1f8WS99F9GORWP_m74l_JfsTHgCrHkbEorHYu72D4Xag/edit) announcement. I can't find the relevant engineers on GitHub to CC: them, so I'll email them a link to this PR as a heads-up.

There's a lot of extra background information that could probably be included, but I think that will be best covered in the article that @surma will eventually write.

Live staged version at https://jeffy.info/samples/css-will-change-transform-rasterization/

### DIFF
--- a/css-will-change-transform-rasterization/demo.js
+++ b/css-will-change-transform-rasterization/demo.js
@@ -1,0 +1,9 @@
+window.addEventListener('load', function() {
+  requestAnimationFrame(function() {
+    var remainsBlurry = document.querySelector('#remainsBlurry');
+    remainsBlurry.style.transform = 'scale(2, 2) translate3d(50px, 0, 0)';
+
+    var noLongerBlurry = document.querySelector('#noLongerBlurry');
+    noLongerBlurry.style.transform = 'scale(2, 2) translate3d(50px, 0, 0)';
+  });
+});
--- a/css-will-change-transform-rasterization/index.html
+++ b/css-will-change-transform-rasterization/index.html
@@ -1,0 +1,58 @@
+---
+feature_name: "Rasterization &amp; will-change: transform"
+chrome_version: 53
+feature_id: 5637351992721408
+---
+
+<h3>Background</h3>
+<p>
+  All content that does not have the <code>will-change: transform</code> CSS
+  property will be re-rastered when its transform scale changes. In other words,
+  <code>will-change: transform</code> effectively means "please animate it fast"
+  without taking the additional time for rasterization. This only applies to
+  scaling that happen via JavaScript manipulation, and does not apply to CSS
+  animations.
+</p>
+
+<h3>Live Output</h3>
+<p>
+  You can see the impact in the sample text below:
+</p>
+
+{% capture html %}
+<div id="container">
+  <div id="remainsBlurry">always blurry</div>
+  <div id="noLongerBlurry">blurry before Chrome 53</div>
+  <div id="alwaysCrisp">always crisp</div>
+</div>
+{% endcapture %}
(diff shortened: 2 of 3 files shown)


css-will-change-transform-rasterization/index.html
```diff
@@ -0,0 +1,58 @@
...
+feature_name: "Rasterization &amp; will-change: transform"
+chrome_version: 53
+feature_id: 5637351992721408
+---
+
+<h3>Background</h3>
+<p>
+  All content that does not have the <code>will-change: transform</code> CSS
+  property will be re-rastered when its transform scale changes. In other words,
+  <code>will-change: transform</code> effectively means "please animate it fast"
```
ebidel: "please animate it fast" is a bit confusing because the demo doesn't actually show an animation. It shows a transform with this before/after behavior.

ebidel:
LGTM

css-will-change-transform-rasterization/index.html
```diff
@@ -0,0 +1,58 @@
...
+<div id="container">
+  <div id="remainsBlurry">always blurry</div>
+  <div id="noLongerBlurry">blurry before Chrome 53</div>
+  <div id="alwaysCrisp">always crisp</div>
+</div>
+{% endcapture %}
+{% include html_snippet.html html=html %}
+
+{% capture css %}
+#container {
```
beaufortfrancois: Can you remove this rule from the CSS that user sees as it doesn't add any value?
Simply add it in an inline `<style>` tag.

css-will-change-transform-rasterization/index.html
```diff
@@ -0,0 +1,58 @@
...
+  margin-left: 50px;
+  margin-top: 100px;
+}
+
+#remainsBlurry {
+  height: 50px;
```
beaufortfrancois: Also remove `height: 50px` from CSS that user sees.

``` css
#container div {
  height: 50px
}

```

css-will-change-transform-rasterization/demo.js
```diff
@@ -0,0 +1,9 @@
+window.addEventListener('load', function() {
```
beaufortfrancois, line 1: Why the `load` event there?

css-will-change-transform-rasterization/demo.js
```diff
@@ -0,0 +1,9 @@
...
+  requestAnimationFrame(function() {
```
beaufortfrancois, line 2: Is `raf` changing the behaviour there or is it more a best practice?

surma, line 2: I _think_ this ensures that rasterization has happened (otherwise we might not see the blurriness)

jeffposnick, line 2: Yeah, this was taken from the engineer's samples and I believe it's required as a best practice.

css-will-change-transform-rasterization/demo.js
jeffposnick, line 1: It's required or else the timing is such that the transformed text won't be blurry. (The logic kicks in for transformations that happen after the page is loaded?)

jeffposnick:
Feedback addressed. Thanks!

PR #399, opened 2016-08-03
[web-bluetooth] Read simply basic USB Vendor Name

beaufortfrancois:
TBR: @jeffposnick 
cc @jyasskin @g-ortuno @scheib


jeffposnick:
It would be nice if you included some information about where the vendor list was sourced from, just in case folks need to look it up directly outside the context of your samples.

scheib:
LGTM, +1 cite source.

beaufortfrancois:
Noted and addressed! See https://github.com/GoogleChrome/samples/pull/403

PR #402, opened 2016-08-03
Basic service worker sample revamp

### LINKED ISSUE #391: Service worker sample for single page static site
I always want the most basic sample for a service worker.

The workers for airhorner and mojibrush are good examples. They gotta do clients.claim() and skipWaiting(), toss files in the cache, and manage those. And that's it.

I'd love to have it here in this repo with the others.  
@addyosmani said he might look into it.

jeffposnick:
R: @addyosmani @paulirish 
CC: @samccone

Staged at https://jeffy.info/samples/service-worker/basic/

This is a revamp of what we're calling the "basic" service worker sample.

I tried to keep the service worker code minimal, and am using a somewhat contrived example of loading in images to simulate how a SPA might handle runtime requests.

You'll notice that there are still a lot of caveats about what the sample does and doesn't do, as I feel like we need to be upfront about the limitations of using this approach, and let developers decide if it's right for them.

Closes #391


paulirish:
This is awesome and I really love the what-it-does/doesn't-do sections. So good.

service-worker/basic/service-worker.js
```diff
@@ -1,12 +1,76 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
 
-  var salutation = 'Hello, ';
-  var whom = decodeURIComponent(event.request.url.match(/\/([^/]*)$/)[1]);
-  var energyLevel = (whom === 'Cleveland') ? '!!!' : '!';
-  var version = '\n\n(Version 1)';
+const CACHES = {
```
paulirish: Feels weird to have a `CACHES` as it feels conflicty with `window.caches`. Makes it harder to search for each. 

Could we rename or just use two basic string variables instead?

service-worker/basic/demo.js
```diff
@@ -0,0 +1,11 @@
+
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.register('service-worker.js');
+}
+
+document.querySelector('#show').addEventListener('click', () => {
```
beaufortfrancois, line 6: Shall we use `_ => {` instead of `() => {` there?

paulirish:
Ace work on this. lgtm % nits

service-worker/basic/demo.js
jeffposnick, line 6: My personal preference is very much for `()`, which matches the style used [elsewhere in this repo](https://github.com/GoogleChrome/samples/blob/gh-pages/arrows-es6/demo.js).

service-worker/basic/service-worker.js
jeffposnick: Changed to two strings.

jeffposnick:
Merging. Thanks for the review and the motivation to revamp the sample!

paulirish:
Most good. :D

PR #405, opened 2016-08-16
Update web payments samples.

rsolomakhin:
- Use a merchant ID and Stripe publishable key in Android Pay demo.
- Simplify feature detection.
- Listen to 'shippingaddresschange' events when requesting shipping
  information.
- Print 'languageCode' of the shipping address.

### DIFF
--- a/paymentrequest/android-pay/demo.js
+++ b/paymentrequest/android-pay/demo.js
@@ -1,19 +1,27 @@
 /**
- * Invokes PaymentRequest for Android Pay.
+ * Invokes PaymentRequest for Android Pay. If you encounter issues when running
+ * your own copy of this sample, run 'adb logcat | grep Wallet' to see detailed
+ * error messages.
  */
 function onBuyClicked() {
   var supportedInstruments = [{
     supportedMethods: ['https://android.com/pay'],
     data: {
-      environment: 'TEST',
-      merchantId: '123456',
+      merchantName: 'Android Pay Demo',
+      // Place your own Android Pay merchant ID here. The merchant ID is tied to
+      // the origin of the website.
+      merchantId: '00184145120947117657',
+      // If you do not yet have a merchant ID, uncomment the following line.
+      // environment: 'TEST',
       allowedCardNetworks: ['AMEX', 'DISCOVER', 'MASTERCARD', 'VISA'],
       paymentMethodTokenizationParameters: {
         tokenizationType: 'GATEWAY_TOKEN',
         parameters: {
           'gateway': 'stripe',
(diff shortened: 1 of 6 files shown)


paymentrequest/free-shipping/demo.js
```diff
@@ -33,8 +33,15 @@ function onBuyClicked() {
 
   var options = {requestShipping: true};
 
-  new PaymentRequest(supportedInstruments, details, options) // eslint-disable-line no-undef
-      .show()
+  var request = new PaymentRequest(supportedInstruments, details, options); // eslint-disable-line no-undef
+
+  request.addEventListener('shippingaddresschange', function(evt) {
+    evt.updateWith(new Promise(function(resolve) {
```
jeffposnick: Can you just use `Promise.resolve(details)` here and elsewhere instead of constructing a new `Promise`?

See http://stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it

paymentrequest/android-pay/demo.js
```diff
@@ -5,15 +5,15 @@ function onBuyClicked() {
   var supportedInstruments = [{
     supportedMethods: ['https://android.com/pay'],
     data: {
-      environment: 'TEST',
-      merchantId: '123456',
+      merchantName: 'Android Pay Demo',
+      merchantId: '00184145120947117657',
```
jeffposnick, line 13: I'm sure you've already thought of this, but I'd feel negligent not to check: is there any risk of abuse if you hardcode this `merchantId`? The same thing goes for the `stripe:publishableKey` value later on.

Abuse aside, what's the experience like if a developer naively copies these keys, not realizing that they need to register their own? This is probably explained elsewhere already, but can you add in a comment in the JavaScript source code stating that they're not meant to be reused, and providing a URL that developers can go to to register their own keys?

rsolomakhin, line 13: Added the appropriate comments and notes on abuse.

Merchant ID cannot be abused, because it's tied to the origin of the website. This merchant ID works only on googlechrome.github.io, rsolomakhin.github.io, and emerald-eon.appspot.com.

Stripe publishable key (pk__) cannot be abused, because one needs the corresponding Stripe secret key (sk__) to initiate a transaction on the server.

There's no public URL for registering a merchant ID yet. I will add it once it goes live.

paymentrequest/free-shipping/demo.js
rsolomakhin: Done.

rsolomakhin:
Addresses comments. PTAL.

jeffposnick:
👍

PR #407, opened 2016-08-22
Improve sw

beaufortfrancois:
This patch is about improving SW Web Bluetooth readability.

R: @jeffposnick

### DIFF
--- a/web-bluetooth/service-worker.js
+++ b/web-bluetooth/service-worker.js
@@ -1,27 +1,24 @@
 function addToCache(request, networkResponse) {
   return caches.open('web-bluetooth')
-    .then((cache) => cache.put(request, networkResponse.clone()));
+    .then(cache => cache.put(request, networkResponse.clone()));
 }
 
 function getCacheResponse(request) {
-  return caches.open('web-bluetooth').then((cache) => {
+  return caches.open('web-bluetooth').then(cache => {
     return cache.match(request);
   });
 }
 
 function getNetworkOrCacheResponse(request) {
-  return new Promise((resolve) => {
-    fetch(request).then((networkResponse) => {
-      addToCache(request, networkResponse);
-      resolve(networkResponse);
-    }).catch(() => {
-      getCacheResponse(request).then((cacheResponse) => {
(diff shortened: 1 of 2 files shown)


web-bluetooth/service-worker.js
```diff
@@ -1,20 +1,27 @@
...
+  });
+}
 
-  event.respondWith(caches.open('web-bluetooth').then(cache => {
-    return cache.match(request).then(response => {
-      var fetchPromise = fetch(request).then(networkResponse => {
-        cache.put(request, networkResponse.clone());
-        return networkResponse;
+function getNetworkOrCacheResponse(request) {
+  return new Promise((resolve) => {
```
jeffposnick: See https://twitter.com/domenic/status/765955660026707969 for a discussion as to why using `new Promise()` like that is an anti-pattern. (I assume this code was carried over from @philipwalton's original post, as he's since updated it.)

web-bluetooth/service-worker.js
```diff
@@ -1,20 +1,27 @@
...
+  return caches.open('web-bluetooth')
+    .then((cache) => cache.put(request, networkResponse.clone()));
+}
 
-  // Don't cache anything that is not on this origin.
-  if (url.origin !== location.origin) {
-    return;
-  }
+function getCacheResponse(request) {
+  return caches.open('web-bluetooth').then((cache) => {
```
jeffposnick: We've been (fairly, I think, for some definition of fairly) consistent about not putting `()` around single-parameter arrow functions in GoogleChrome/samples.

Could you rewrite this to `then(cache => {`, and make that change elsewhere as well?

(I need to get a rule for http://eslint.org/docs/rules/arrow-parens added to our standard `eslint` config.)

web-bluetooth/service-worker.js
beaufortfrancois: Best review feedback ever ;)
Thank you @jeffposnick. I'll update code.

beaufortfrancois:
Thank you @jeffposnick 
Feedback is addressed.

jeffposnick:
👍

PR #408, opened 2016-08-23
Add eslint rule about single-parameter arrow functions.

beaufortfrancois:
As mentioned in https://github.com/GoogleChrome/samples/pull/407#discussion_r75683772 by @jeffposnick , here's a eslint rule about single-parameter arrow functions.


jeffposnick:
Thanks! 👍

PR #409, opened 2016-08-29
Update PaymentRequest Chrome version.

rsolomakhin:
Chrome 53 is the first version where PaymentRequest is enabled by
default.


ebidel:
LGTM

PR #410, opened 2016-08-31
Writing to a characteristic does not neccesarily mean that subsiquent…

beaufortfrancois:
… reads will return that value

I'll push this CL when https://github.com/WebBluetoothCG/web-bluetooth/issues/277 is merged.


jeffposnick:
👍

PR #414, opened 2016-09-08
Update to the window.caches sample

### LINKED ISSUE #413: Update service-worker/window-caches to test for 'caches' in window

jeffposnick:
R: @jmadler 

Modernizes the sample to reflect some changes to the cache.add() behavior, as well as moving to the new templating system. It also explicitly checks for `window.caches` and displays an error status when it's not available.

Fixes #413


jeffposnick:
(This is staged at https://jeffy.info/samples/service-worker/window-caches/)

jmadler:
LGTM

PR #415, opened 2016-09-08
Custom offline page: avoid cache-busting "regular" requests

ithinkihaveacat:
The request for `OFFLINE_URL` should be cache-busted to avoid surprises, but this isn't needed for regular requests.


jeffposnick:
👍 Thanks!

PR #416, opened 2016-09-09
Foreign fetch sample

jeffposnick:
R: @addyosmani @jakearchibald @ebidel @PaulKinlan @beaufortfrancois etc.
CC: @mkruisselbrink

Here's a basic sample illustrating the use of foreign fetch to handle offline use cases for a trivial random-number API.

The code for the service that's checked in here is a copy of what's currently deployed at https://foreign-fetch-demo.appspot.com

Feel free to review at this time, but **DO NOT MERGE** yet, as there's a [serious bug](https://bugs.chromium.org/p/chromium/issues/detail?id=645281) in the current M54 implementation of foreign fetch that will cause the `fetch()` in the service worker to loop for ~10 seconds then fail.

The https://developers.google.com/web/updates/2016/09/foreign-fetch URL referenced in this post also is not live yet (waiting on the same bug to be resovled).

### DIFF
--- a/service-worker/foreign-fetch/foreign-fetch-sw.js
+++ b/service-worker/foreign-fetch/foreign-fetch-sw.js
@@ -1,0 +1,21 @@
+// This is deployed to https://foreign-fetch-demo.appspot.com/foreign-fetch-sw.js
+
+self.addEventListener('install', event => {
+  event.registerForeignFetch({
+    scopes: ['/random'], // or self.registration.scope to handle everything
+    origins: ['*'] // or ['https://client1.com'] to limit the remote origins
+  });
+});
+
+self.addEventListener('foreignfetch', event => {
+  event.respondWith(
+    fetch(event.request) // Try to make a network request
+      .catch(() => new Response('34')) // Offline? Your random number is 34!
+      .then(response => {
+        return {
+          response,
+          origin: event.origin // Make this a CORS response
+        };
+      })
+  );
+});
--- a/service-worker/foreign-fetch/foreign-fetch-sw.js
+++ b/service-worker/foreign-fetch/foreign-fetch-sw.js
@@ -3,7 +3,7 @@
 self.addEventListener('install', event => {
   event.registerForeignFetch({
     scopes: ['/random'], // or self.registration.scope to handle everything
-    origins: ['*'] // or ['https://client1.com'] to limit the remote origins
+    origins: ['*'] // or ['https://example.com'] to limit the remote origins
   });
 });
 
--- a/service-worker/foreign-fetch/remote-server.js
+++ b/service-worker/foreign-fetch/remote-server.js
@@ -1,0 +1,38 @@
+// This is the web server implementation for the "random number API".
+// It is currently deployed to https://foreign-fetch-demo.appspot.com
+
+const express = require('express');
+const app = express();
+
+const SW_JS_FILE = 'foreign-fetch-sw.js';
+const MAX_RANDOM_NUMBER = 100;
+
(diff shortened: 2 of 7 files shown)


addyosmani:
Thanks, Jeff. It looks like the fix for the issue referenced landed a few hours ago in https://crrev.com/65299f214fad8f609c5872bcdd5f4ea82ec29713 and should be in latest trunk/Canary in a day or two. Fwiw: the deployed version of the demo in this PR looks like it may be down atm, but will check again later.

service-worker/foreign-fetch/foreign-fetch-sw.js
```diff
@@ -0,0 +1,21 @@
+// This is deployed to https://foreign-fetch-demo.appspot.com/foreign-fetch-sw.js
+
+self.addEventListener('install', event => {
+  event.registerForeignFetch({
+    scopes: ['/random'], // or self.registration.scope to handle everything
+    origins: ['*'] // or ['https://client1.com'] to limit the remote origins
```
addyosmani: I would replace `https://client1.com` with `<client-1 origin-domain>` or something that isn't tied to a domain that might exist (there doesn't appear to be anything at client1.com right now in case anyone was wondering :P)

service-worker/foreign-fetch/remote-server.js
```diff
@@ -0,0 +1,38 @@
+// This is the web server implementation for the "random number API".
+// It is currently deployed to https://foreign-fetch-demo.appspot.com
+
+const express = require('express');
```
addyosmani, line 4: We're including a remote server example here but not a package.json file for folks to grab the dependency it uses. Worth adding?

jeffposnick:
The remote "API" service is hosted on https://foreign-fetch-demo.appspot.com/random, and the FF SW is served from https://foreign-fetch-demo.appspot.com/foreign-fetch-sw.js. Those both look up. But there's (deliberately) nothing handling the root of https://foreign-fetch-demo.appspot.com, so maybe that's what you saw.

I don't have the client-side code staged anywhere, partly because you'll trigger an infinite loop if you make use of it...

(We don't have confirmation yet as to when the fix will be merged from Canary into Beta.)

service-worker/foreign-fetch/remote-server.js
jeffposnick: Sure, I'll add for the benefit of folks who are poking around the source tree. Though I'm assuming that most folks will see `remote-server.js` inline in the templatized web page.

service-worker/foreign-fetch/foreign-fetch-sw.js
jakearchibald: `example.com` is reserved for examples

jeffposnick:
Updated with review feedback, and it's now deployed to https://jeffy.info/samples/service-worker/foreign-fetch/

The "API" server has also been updated to include a real Origin Trials token.

It should work as expected in Chrome Canary, but the [fix](https://bugs.chromium.org/p/chromium/issues/detail?id=645281) for the infinite loop bug hasn't been merged to Chrome Beta at this time.

service-worker/foreign-fetch/remote-server.js
jeffposnick: Done in the latest commits.

jeffposnick:
Given that the bug fix is tagged to be merged into the next M54 build, and given that getting this submitted is a prerequisite for publishing the /web/updates article, I'm going to merge this but not make any noise about it until the article comes out.

Happy to file a follow-up PR with any updates after the fact.

PR #417, opened 2016-09-13
Remove the 'View Source' link from the template

### LINKED ISSUE #294: "View Source" link (in Fallback Response Sample) doesn't work in Firefox
Security Error: Content at https://googlechrome.github.io/samples/service-worker/fallback-response/index.html may not load or link to view-source:https://googlechrome.github.io/samples/service-worker/fallback-response/.

jeffposnick:
R: @beaufortfrancois @addyosmani @ebidel etc.

Chrome 53 has followed the lead of other browsers, and no longer supports links with the `view-source:` scheme. There's no reason to keep it in the template.

The link to view the original source code on GitHub should suffice, and I'm assuming that most developers who are properly motivated will just right-click and choose "View Source"/use the DevTools.

Closes #294


beaufortfrancois:
Lgtm

PR #419, opened 2016-09-15
Removed deprecated 'gcm_user_visible_only'

karolklp:
According to ChromeStatus this key is already deprecated (https://www.chromestatus.com/feature/5778950739460096). Proper code is already put in https://github.com/GoogleChrome/samples/blob/gh-pages/push-messaging-and-notifications/main.js : 119


jeffposnick:
Thanks, @karolklp!

PR #424, opened 2016-10-07
Add ResizeObserver sample

surma:
cc @atotic as an FYI

### DIFF
--- a/resizeobserver/demo.js
+++ b/resizeobserver/demo.js
@@ -1,0 +1,13 @@
+/* global ResizeObserver */
+
+const ro = new ResizeObserver(entries => {
+  for (let entry of entries) {
+    if (entry.contentRect.width < 250) {
+      entry.target.style.border = '3px solid red';
+    } else {
+      entry.target.style.border = 'none';
+    }
+  }
+});
+// Only observe the green box (which is the third box).
+ro.observe(document.querySelector('.box:nth-child(3)'));
--- a/resizeobserver/demo.js
+++ b/resizeobserver/demo.js
@@ -2,8 +2,8 @@
 
 const ro = new ResizeObserver(entries => {
   for (let entry of entries) {
-    entry.target.classList.toggle('stripes', entry.contentRect.width < 250);
+    entry.target.style.borderRadius = Math.max(0, 250 - entry.contentRect.width) + 'px';
   }
 });
-// Only observe the green box (which is the third box).
+// Only observe the 2nd box
 ro.observe(document.querySelector('.box:nth-child(2)'));
--- a/resizeobserver/demo.js
+++ b/resizeobserver/demo.js
@@ -2,12 +2,8 @@
 
 const ro = new ResizeObserver(entries => {
   for (let entry of entries) {
-    if (entry.contentRect.width < 250) {
-      entry.target.style.border = '3px solid red';
-    } else {
-      entry.target.style.border = 'none';
-    }
+    entry.target.classList.toggle('stripes', entry.contentRect.width < 250);
   }
 });
 // Only observe the green box (which is the third box).
-ro.observe(document.querySelector('.box:nth-child(3)'));
+ro.observe(document.querySelector('.box:nth-child(2)'));
--- a/resizeobserver/index.html
+++ b/resizeobserver/index.html
@@ -15,7 +15,7 @@ feature_id: 5705346022637568
   -like behavior on a per-element basis.
 <p>
   The <a href="https://developers.google.com/web/updates/2016/10/resizeobserver"><em>Google Developers Web Updates</em></a>
-  article provides more details on <code>ResizeObserver</code>.
+  article provides more details on <code>ResizeObserver</code>. As of Chrome 54, it is behind the `chrome://flags/#enable-experimental-web-platform-features` flag.
 </p>
 
 {% capture initial_output_content %}
@@ -27,24 +27,15 @@ feature_id: 5705346022637568
 .boxes {
   display: flex;
   flex-wrap: wrap;
+  width: 40vw;
 }
 .box {
   height: 200px;
-  flex-basis: 200px;
   flex-shrink: 1;
   flex-grow: 1;
   margin: 10px;
   box-sizing: border-box;
   border: 3px solid black;
-}
-.stripes {
-  background: repeating-linear-gradient(
-    45deg,
-    #fff,
-    #fff 10px,
-    #000 10px,
-    #000 20px
-  );
 }
 </style>
 <div class="boxes">
@@ -53,10 +44,5 @@ feature_id: 5705346022637568
 </div>
 {% endcapture %}
 {% include output_helper.html initial_output_content=initial_output_content %}
-
-{% capture css %}
-
-{% endcapture %}
-{% include css_snippet.html css=css %}
 
 {% include js_snippet.html filename='demo.js' %}
--- a/resizeobserver/index.html
+++ b/resizeobserver/index.html
@@ -1,0 +1,52 @@
+---
+feature_name: ResizeObserver
+chrome_version: 54
+feature_id: 5705346022637568
+---
+
+<h3>Background</h3>
+<p>
+  The <a href="https://wicg.github.io/ResizeObserver/">ResizeObserver API</a>
+  provides an API to get notified whenever an element changes its size.
+</p>
+<p>
+  This sample illustrates how to use <code>ResizeObserver</code> to implement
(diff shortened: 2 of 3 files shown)


resizeobserver/index.html
```diff
@@ -0,0 +1,52 @@
...
+---
+
+<h3>Background</h3>
+<p>
+  The <a href="https://wicg.github.io/ResizeObserver/">ResizeObserver API</a>
+  provides an API to get notified whenever an element changes its size.
+</p>
+<p>
+  This sample illustrates how to use <code>ResizeObserver</code> to implement
+  media query-like behavior on a per-element basis.
```
jeffposnick: Link to https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries

resizeobserver/demo.js
```diff
@@ -0,0 +1,13 @@
+/* global ResizeObserver */
+
+const ro = new ResizeObserver(entries => {
```
jeffposnick, line 3: This code throws a `ResizeObserver loop limit exceeded` error as soon as the page is opened. (Testing on Chrome 55.0.2883.0 with the "Experimental Web Platform Features" flag enabled.)

image

resizeobserver/index.html
```diff
@@ -0,0 +1,52 @@
...
+<p>
+  The <a href="https://developers.google.com/web/updates/2016/10/resizeobserver"><em>Google Developers Web Updates</em></a>
+  article provides more details on <code>ResizeObserver</code>.
+</p>
+
+{% capture initial_output_content %}
+<p>
+  On browsers that support <code>ResizeObserver</code>, you should see a border
+  appear around the green box when resizing the window.
```
jeffposnick: For folks who have difficulty distinguishing colors, could you rephrase as something like "the third box" rather than "the green box"?

Also, any particular reason why there are six boxes? It seems like you could make your point with just two boxes—one of which is associated with an RO, and the other which isn't.

resizeobserver/index.html
```diff
@@ -0,0 +1,52 @@
...
+  <div class="box" style="background: hsl(60, 100%, 50%)"></div>
+  <div class="box" style="background: hsl(120, 100%, 50%)"></div>
+  <div class="box" style="background: hsl(180, 100%, 50%)"></div>
+  <div class="box" style="background: hsl(240, 100%, 50%)"></div>
+  <div class="box" style="background: hsl(300, 100%, 50%)"></div>
+</div>
+{% endcapture %}
+{% include output_helper.html initial_output_content=initial_output_content %}
+
+{% capture css %}
```
jeffposnick: This could just be inline in a `<style>` element alongside the other HTML. I don't think there's anything important there that developers would need to see echoed to the screen.

resizeobserver/index.html
```diff
@@ -0,0 +1,52 @@
...
+.boxes {
+  display: flex;
+  flex-wrap: wrap;
+}
+.box {
+  height: 200px;
+  flex-basis: 200px;
```
jeffposnick: Our samples template [imposes](https://github.com/GoogleChrome/samples/blob/gh-pages/styles/main.css#L7) a maximum width of `800px` on the `<body>` element when the viewport is wider than `832px`. As a result, the boxes won't change width if the window is resized until it gets below 832px.

jeffposnick (changes requested):

resizeobserver/demo.js
atotic: That's because border is being changed in observer callback, which triggers another resize event immediately.
This is not a great demonstration of RO. RO's purpose to change the Element's content in response to resize, not to change the element being resized.

resizeobserver/index.html
surma: Good point about the colors. The only reason it’s 6 boxes because it’s the exact same example from my /updates article. I’ll adjust accordingly

resizeobserver/demo.js
surma: I switched to changing the background image. WDYT @atotic? This sample is more about showing usage and I wanted to keep it small. I’ll rework the sample for the `/updates` article.

screenshot 2016-10-08 09 43 22

atotic: This code is not buggy. But, it does not really show what RO is about. If I was a developer, my reaction would be "huh?". If we are set on two rectangles, how about making one of them a circle through border-radius, and then maintain the circleness by changing the border radius. If I saw that, I'd be like "neat!".

surma: I can’t quite follow. Changing the background pattern is just as much a (useless) visual change as changing the element to a circle. I am trying to show _how_ to use RO, which this easily conveys (and can be tested by the user). What am I missing?

atotic: It is my personal preference. These are my reasons:
- border-radius is continuous. You see circle changing smoothly every time you resize the element.
- background-image is discrete. Nothing changes unless you cross the 250px threshold. If the user is not resizing the window in the right direction, he might miss the change. Also, are we trying to make the demo phone-friendly?
- circle border that stays circle is something I'd have a very hard time coding in JS today. Seeing RO make it easy makes me a happy developer. Changing background-image based on size feels doable thought some @ rules hack.

surma: Ah! I now I understand. Thanks for clarifying. The continuous bit is indeed good to show off. Will change :)

surma:
@jeffposnick I think this is good to go from our side

resizeobserver/index.html
jeffposnick: Now that you've moved the styles out, you should get rid of the `{% capture css %}` and corresponding `{% include ... %}`, to avoid the empty section:

image

resizeobserver/index.html
```diff
@@ -0,0 +1,62 @@
...
+</p>
+<p>
+  This sample illustrates how to use <code>ResizeObserver</code> to implement
+  [media query](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries)
+  -like behavior on a per-element basis.
+<p>
+  The <a href="https://developers.google.com/web/updates/2016/10/resizeobserver"><em>Google Developers Web Updates</em></a>
+  article provides more details on <code>ResizeObserver</code>.
+</p>
+
```
jeffposnick, line 20: You need a section here letting developers know that as of right now, the feature requires `chrome://flags/#enable-experimental-web-platform-features`

resizeobserver/index.html
```diff
@@ -0,0 +1,62 @@
...
+.box {
+  height: 200px;
+  flex-basis: 200px;
+  flex-shrink: 1;
+  flex-grow: 1;
+  margin: 10px;
+  box-sizing: border-box;
+  border: 3px solid black;
+}
+.stripes {
```
jeffposnick: `.stripes` isn't used anymore, right?

resizeobserver/index.html
```diff
@@ -0,0 +1,62 @@
...
+  article provides more details on <code>ResizeObserver</code>.
+</p>
+
+{% capture initial_output_content %}
+<p>
+  On browsers that support <code>ResizeObserver</code>, you should see the border
+  radius of the second box change when resizing the window.
+</p>
+<style>
+.boxes {
```
jeffposnick, line 27: You've still got the aforementioned problem where the maximum size imposed on the `<body>` element by the default template's styles leads to a situation in which this container won't actually change size until the viewport gets below 800px or so. I.e. if your initial viewport is 1000px, and you resize the viewport to 900px, nothing happens, defeating the point of this sample.

Here's a suggestion for changing the styles so that the size of this container always takes the viewport width into account, so that any resizing will lead to the `ResizeObserver` triggering:

``` css
.boxes {
  display: flex;
  flex-wrap: wrap;
  width: 40vw;
}
.box {
  height: 200px;
  flex-shrink: 1;
  flex-grow: 1;
  margin: 10px;
  box-sizing: border-box;
  border: 3px solid black;
}
```

(It adds a `width` to `.boxes`, and removes the `flex-basis` from `.box`.)

jeffposnick (changes requested):

jeffposnick:
@surma there were a few earlier review comments that still need to be addressed (e.g. https://github.com/GoogleChrome/samples/pull/424#discussion_r82468593) and then a few new ones. Thanks!

surma:
@jeffposnick Thanks for keeping an eye out. Apparently I was half asleep when I fixed this. Implemented all your things (hopefully).

jeffposnick (approved):

PR #426, opened 2016-10-10
bluetooth: Add Battery Level ES7 Async sample

beaufortfrancois:
With ES7 Async functions (async await) in M55 now, I thought we may want to add a sample on how to use them with the Web Bluetooth API.

@jeffposnick @scheib @jyasskin @g-ortuno

See https://beaufortfrancois.github.io/samples/web-bluetooth/battery-level-async-await.html and https://beaufortfrancois.github.io/samples/web-bluetooth/index.html

screenshot 2016-10-10 at 2 52 10 pm

### DIFF
--- a/web-bluetooth/battery-level-async-await.js
+++ b/web-bluetooth/battery-level-async-await.js
@@ -17,7 +17,7 @@ async function onButtonClick() {
     const value = await characteristic.readValue();
 
     log('> Battery Level is ' + value.getUint8(0) + '%');
-  } catch(e) {
+  } catch(error) {
     log('Argh! ' + error);
   }
 }
--- a/web-bluetooth/battery-level-async-await.js
+++ b/web-bluetooth/battery-level-async-await.js
@@ -1,0 +1,23 @@
+async function onButtonClick() {
+  try {
+    log('Requesting Bluetooth Device...');
+    const device = await navigator.bluetooth.requestDevice({
+        filters: [{services: ['battery_service']}]});
+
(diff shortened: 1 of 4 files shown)


jeffposnick:
The only feedback that I have is that ES7 isn't a meaningful term. It's colloquially used to describe anything that's due to be standardized but didn't make it into ES2015, but given its lack of precision, I'd leave it out.

I'd just refer to "async/await" without the "ES7" bit, and include a link to an article describing what it's all about. (@jakearchibald has one due to be published on Web Fundamentals, once the M55 beta ships.)

web-bluetooth/battery-level-async-await.js
```diff
@@ -0,0 +1,23 @@
...
+    const service = await server.getPrimaryService('battery_service');
+
+    log('Getting Battery Level Characteristic...');
+    const characteristic = await service.getCharacteristic('battery_level');
+
+    log('Reading Battery Level...');
+    const value = await characteristic.readValue();
+
+    log('> Battery Level is ' + value.getUint8(0) + '%');
+  } catch(e) {
```
scheib: catch(error)

scheib (changes requested):

web-bluetooth/battery-level-async-await.js
beaufortfrancois: Oops. Thanks.

beaufortfrancois:
I've addressed all your feedback.
I'll add jake's article when it is up. Thanks!

PR #431, opened 2016-10-18
Add more Web Bluetooth Async Await samples

beaufortfrancois:
Here's a patch that add a "Async Await" edition to all our Web Bluetooth samples. I believe this will make developers happy when they see how easy it is to use.

You can find all updated samples at https://beaufortfrancois.github.io/samples/web-bluetooth/index.html

R: @jeffposnick @scheib @jyasskin


web-bluetooth/index.html
```diff
@@ -13,22 +13,22 @@
...
 
 <h3>Combining multiple operations</h3>
 
-<p><a href="gap-characteristics.html">GAP Characteristics</a> - get all GAP characteristics of a BLE Device.</p>
-<p><a href="device-information-characteristics.html">Device Information Characteristics</a> - get all Device Information characteristics of a BLE Device.</p>
-<p><a href="link-loss.html">Link Loss</a> - set the Alert Level characteristic of a BLE Device (readValue & writeValue).</p>
-<p><a href="discover-services-and-characteristics.html">Discover Services & Characteristics</a> - discover all accessible primary services and their characteristics from a BLE Device.</p>
-<p><a href="automatic-reconnect.html">Automatic Reconnect</a> - Reconnect to a disconnected BLE device using an exponential backoff algorithm.</p>
-<p><a href="read-characteristic-value-changed.html">Read Characteristic Value Changed</a> - read battery level and be notified of changes from a BLE Device.</p>
+<p><a href="gap-characteristics.html">GAP Characteristics (Promises)</a> / <a href="gap-characteristics.html">GAP Characteristics (Async Await)</a> - get all GAP characteristics of a BLE Device.</p>
```
scheib: async link needs "-async-await"

scheib (commented):
I'm eager for async-await versions as well, but I'm not sure we should have parallel sets of all these samples. I see on https://www.chromestatus.com/feature/5643236399906816 that the feature is shipping in 55 ~Dec 6th stable, so perhaps it's OK to have the blend of both for a shorter period of time. After 55 is in stable, do you plan on removing the promise versions?

jyasskin:
I'd be sad to leave Servo behind. @dati91 said they weren't going to have async/await soon, so we should probably keep the non-await versions until they catch up. We could maybe include instructions for compiling the await versions to Servo-compatible code, to get the best of both worlds?

web-bluetooth/index.html
beaufortfrancois: Done.

beaufortfrancois:
I was planning on keeping both (Promises and Async/Await) versions until Async/Await ships on all modern browsers (including Servo).

scheib (approved):

PR #432, opened 2016-10-18
Show error in console

beaufortfrancois:
When there's an error in our code, it is hard to know where it is exactly because of the global error handler. This patch simply shows it in JS Console.

R: @jeffposnick


jeffposnick:
👍

PR #439, opened 2016-11-08
Remove 'careOf' field from PaymentAddress.

rsolomakhin:
The 'careOf' field has been removed from PaymentAddress.


jeffposnick:
👍

PR #444, opened 2016-11-17
Override page index.html for web bluetooth samples

beaufortfrancois:
Hello @jeffposnick, this patch gives the ability for individual samples to override their page index.html so that "Browse Samples" top link is unique.

For instance, all individual Web Bluetooth samples now use https://googlechrome.github.io/samples/web-bluetooth/index.html instead of https://www.chromestatus.com/samples. 
https://googlechrome.github.io/samples/web-bluetooth/index.html still use https://www.chromestatus.com/samples though.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -53,7 +53,7 @@ limitations under the License.
     <p class="availability">
       Available in <a target="_blank" href="https://www.chromestatus.com/feature/{{ page.feature_id }}">Chrome {{ page.chrome_version }}+</a> |
       <a target="_blank" href="{{ site.gh_url_prefix }}{{ page.url | remove_first: 'index.html' }}">View on GitHub</a> |
-      <a target="_blank" href="https://www.chromestatus.com/samples">Browse Samples</a>
+      <a {% if page.index == null %}target="_blank"{% endif %} href="{% if page.index %}{{ page.index }}{% else %}https://www.chromestatus.com/samples{% endif %}">Browse Samples</a>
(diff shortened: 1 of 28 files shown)


_layouts/default.html
```diff
@@ -53,7 +53,7 @@
     <p class="availability">
       Available in <a target="_blank" href="https://www.chromestatus.com/feature/{{ page.feature_id }}">Chrome {{ page.chrome_version }}+</a> |
       <a target="_blank" href="{{ site.gh_url_prefix }}{{ page.url | remove_first: 'index.html' }}">View on GitHub</a> |
-      <a target="_blank" href="https://www.chromestatus.com/samples">Browse Samples</a>
+      <a {% if page.index %}{% else %}target="_blank"{% endif %} href="{% if page.index %}{{ page.index }}{% else %}https://www.chromestatus.com/samples{% endif %}">Browse Samples</a>
```
jeffposnick: Instead of relying on the `{% else %}` clause, can you switch this to `{% if !page.index %}` (or whatever the equivalent in Liquid templates is)?

beaufortfrancois: There is no "NOT" operator in Liquid templates. https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#if--else

Here's a new patch: https://github.com/GoogleChrome/samples/pull/444/commits/227573a116d919f05735d114377958016519fd40 WDYT?

jeffposnick:
👍

PR #452, opened 2016-11-25
Adds WebVR + Three.js sample

paullewis:
Adds a Hello World Three.js sample for WebVR. Most of the interesting stuff is in the demo-vr.js file :)

PTAL @toji @wibblymat @surma


web-vr/hello-world/demo/demo-vr.js
```diff
@@ -0,0 +1,203 @@
...
+      frameData: new VRFrameData()
+    };
+
+    this._addVREventListeners();
+    this._getDisplays();
+  }
+
+  _addVREventListeners () {
+    window.addEventListener('vrdisplayactivate', evt => {
+      evt.preventDefault();
```
surma, line 45: Is the `preventDefault()` actually still necessary?

surma (commented):
One Q, otherwise LGTM.

web-vr/hello-world/demo/demo-vr.js
```diff
@@ -0,0 +1,203 @@
...
+        'Disable VR' :
+        'Enable VR';
+  }
+
+  _deactivateVR () {
+    if (!this._vr.display) {
+      return;
+    }
+
+    this._vr.display.exitPresent();
```
wibblymat, line 103: 1. You return a `Promise` from this method, but I don't know why
1. `activateVR` does NOT return a `Promise`, even though it's similar
1. `exitPresent` and `requestPresent` return `Promise`s, if you really want to return one.
1. Caveat - `exitPresent` will reject if you weren't actually presenting

web-vr/hello-world/demo/demo-vr.js
```diff
@@ -0,0 +1,203 @@
...
+      this._scene.matrixAutoUpdate = true;
+
+      return super._render();
+    }
+
+    const EYE_WIDTH = this._width * 0.5;
+    const EYE_HEIGHT = this._height;
+
+    // Get all the latest data from the VR headset and dump it into frameData.
+    this._vr.display.getFrameData(this._vr.frameData);
```
wibblymat, line 146: I don't see a submitFrame call... isn't that needed to make the matrices update? I believe that this is somewhere that the polyfill and extension differ from the spec, so check in a real device.

wibblymat (commented):
Only reviewed demo-vr.js

web-vr/hello-world/demo/demo-vr.js
paullewis: Removed. AFAIK it's a noop.

web-vr/hello-world/demo/demo-vr.js
paullewis: Fixed

web-vr/hello-world/demo/demo-vr.js
paullewis: Fixed

PR #455, opened 2016-12-03
Update engagement period

dandv:


PaulKinlan:
I'm going to merge this, because it is more correct than what we have, we still need to fix it.

PR #458, opened 2016-12-06
bluetooth: Add acceptAllDevices

beaufortfrancois:
You can give a try to this updated samples at https://beaufortfrancois.github.io/samples/web-bluetooth.

https://beaufortfrancois.github.io/samples/web-bluetooth/device-info.html and any sample previously using `anyNamedDevice` have been updated to use `acceptAllDevices`.

R: @g-ortuno @jeffposnick

### DIFF
--- a/web-bluetooth/automatic-reconnect-async-await.js
+++ b/web-bluetooth/automatic-reconnect-async-await.js
@@ -5,7 +5,7 @@ async function onButtonClick() {
   try {
     log('Requesting any Bluetooth Device...');
     bluetoothDevice = await navigator.bluetooth.requestDevice({
-        filters: anyNamedDevice()});
+        acceptAllDevices: true});
     bluetoothDevice.addEventListener('gattserverdisconnected', onDisconnected);
     connect();
   } catch(error) {
@@ -50,14 +50,6 @@ async function exponentialBackoff(max, delay, toTry, success, fail) {
       exponentialBackoff(--max, delay * 2, toTry, success, fail);
     }, delay * 1000);
   }
-}
-
-function anyNamedDevice() {
-  // This is the closest we can get for now to get all devices.
-  // https://github.com/WebBluetoothCG/web-bluetooth/issues/234
-  return Array.from('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ')
-      .map(c => ({namePrefix: c}))
-      .concat({name: ''});
 }
 
 function time(text) {
--- a/web-bluetooth/automatic-reconnect-async-await.js
+++ b/web-bluetooth/automatic-reconnect-async-await.js
@@ -5,6 +5,7 @@ async function onButtonClick() {
   try {
     log('Requesting any Bluetooth Device...');
     bluetoothDevice = await navigator.bluetooth.requestDevice({
+     // filters: [...] <- Prefer filters to save energy & show relevant devices.
         acceptAllDevices: true});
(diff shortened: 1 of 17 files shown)


web-bluetooth/device-info.js
```diff
@@ -19,8 +19,16 @@ function onButtonClick() {
...
+    acceptAllDevices: document.querySelector('#allDevices').checked
+  };
+  if (filters.length) {
+    options.filters = filters;
+  }
+
   log('Requesting Bluetooth Device...');
-  navigator.bluetooth.requestDevice({filters: filters})
+  log('with ' + JSON.stringify(options));
+  navigator.bluetooth.requestDevice(options)
```
g-ortuno, line 31: Note that you are not allowed to use acceptAllDevices with filters so you might need some logic for that.

web-bluetooth/read-characteristic-value-changed.js
```diff
@@ -15,8 +15,8 @@ function onReadBatteryLevelButtonClick() {
 
 function requestDevice() {
   log('Requesting Bluetooth Device...');
-  return navigator.bluetooth.requestDevice(
-    {filters: anyNamedDevice(), optionalServices: ['battery_service']})
+  return navigator.bluetooth.requestDevice({
+      acceptAllDevices: true, optionalServices: ['battery_service']})
```
g-ortuno: It makes me sad that we are using acceptAllDevices so liberally. I recognize the desire to make our samples super easy to use but every developer that copies this code will end up using it which will result in a bunch of unrelated devices being shown in the chooser and energy being wasted because we have no filters.

web-bluetooth/device-info.js
beaufortfrancois: Thanks! I've fixed that in https://github.com/GoogleChrome/samples/pull/458/commits/4ebbcd100891bb6984bc044647d36ddc914bba9b

web-bluetooth/read-characteristic-value-changed.js
beaufortfrancois: Would https://github.com/GoogleChrome/samples/pull/458/commits/9533907bcc3c6d3e34016fc3dd8d45f553e9d160 be enough?

web-bluetooth/automatic-reconnect-async-await.js
```diff
@@ -3,9 +3,9 @@ var bluetoothDevice;
 async function onButtonClick() {
   bluetoothDevice = null;
   try {
-    log('Requesting any Bluetooth Device...');
+    log('Requesting any Bluetooth Device... (NOT RECOMMENDED - USE FILTERS)');
```
g-ortuno: I don't think it's that clear what exactly is not recommended?

```js
navigator.bluetooth.requestDevice({
  acceptAllDevices: true /* NOT RECOMMENDED - USE FILTERS */});
```

Also fwiw: 10% of the devices requests have been using anyNamedDevice().

g-ortuno: Or maybe:
```js
navigator.bluetooth.requestDevice({
  // filters [{...}] <- Prefer to use filters to save energy and show only relevant devices.
  acceptAllDevices: true
});
```

beaufortfrancois: Thanks @g-ortuno! How about https://github.com/GoogleChrome/samples/pull/458/commits/5f0a821b2248564ca8b195adb1e8553a43658cf4?

g-ortuno:
lgtm

beaufortfrancois:
I'll wait for M56 to land in Stable channel before merging this CL then.

PR #472, opened 2017-01-27
[DO NOT MERGE] Very basic Web Assembly demo

wibblymat:
DO NOT MERGE

Wasn't sure what level to aim at here, so I've gone for 'minimally useful'.

Circumvented the usual `js_snippet` include here because I wanted more than one output box.

CC: @jeffposnick @s3ththompson (mostly FYI) @PaulKinlan


wibblymat:
Marked with [DO NOT MERGE] because apparently this will not be on by default at the start of the beta phase, even though it WILL be on by default by 57 *stable*.

Will need to revisit when the flag changes in a couple of weeks.

web-assembly/index.html
```diff
@@ -0,0 +1,63 @@
...
+---
+
+<h3>Background</h3>
+<p>Web Assembly is a binary code format for the web.</p>
+
+<script>
+  function log(id, ...args) {
+    const el = document.getElementById(id);
+
+    if (!el) {
```
beaufortfrancois, line 14: If you move the `log` function at the very bottom, you shouldn't need this check I believe.

web-assembly/index.html
```diff
@@ -0,0 +1,63 @@
...
+{% endcapture %}
+
+{% include js_snippet.html title="" js=compile %}
+
+<h3>Calling exported functions</h3>
+<p>Each instance has an <code>exports</code> property that gives you the public interface of the module. In this example there is only one export, which is a function.</p>
+
+{% capture call %}
+instancePromise.then(instance => {
+  log('call', 'Exports:', Object.keys(instance.exports));
```
beaufortfrancois, line 56: Here's what I get ;(

```
Exports: 
Answer: 42
```

web-assembly/index.html
```diff
@@ -0,0 +1,63 @@
...
+const instancePromise = fetch('demo.wasm')
+  .then(response => response.arrayBuffer())
+  .then(buffer => WebAssembly.compile(buffer))
+  .then(module => new WebAssembly.Instance(module));
+{% endcapture %}
+
+{% include js_snippet.html title="" js=compile %}
+
+<h3>Calling exported functions</h3>
+<p>Each instance has an <code>exports</code> property that gives you the public interface of the module. In this example there is only one export, which is a function.</p>
```
beaufortfrancois, line 52: Could we have a chance to look at `demo.wasm` code?

web-assembly/README.md
```diff
@@ -0,0 +1,5 @@
+Web Assembly Sample
+===
+See https://googlechrome.github.io/samples/web-assembly/index.html for a live demo.
```
beaufortfrancois, line 3: If there are going to more samples, I would recommend doing something like https://googlechrome.github.io/samples/web-bluetooth/ and https://googlechrome.github.io/samples/image-capture/

web-assembly/index.html
```diff
@@ -0,0 +1,63 @@
...
+
+    el.innerHTML += args.join(' ') + '<br/>';
+  }
+</script>
+
+{% capture detect %}
+let wasmSupported = 'WebAssembly' in window;
+
+if (wasmSupported) {
+  log('detect', 'Web Assembly is supported');
```
beaufortfrancois, line 27: `s/supported/supported./`

web-assembly/index.html
beaufortfrancois: For info,

```
Object.keys(instance.exports)
> []
```

beaufortfrancois:
@wibblymat What is the status of this sample?

wibblymat:
@beaufortfrancois This was originally intended to go out along with a beta blog post, but shortly beforehand all mention of WASM was pulled from the post. I had a separate article that linked off to this sample, which was cancelled at the same time.

I think I'm going to close, because this is a very simple example and there are now much better ones available elsewhere.

PR #480, opened 2017-02-02
Add write descriptor sample

beaufortfrancois:
This sample is the one featuring how to write a value to a Bluetooth GATT descriptor.

Try them live at:
https://beaufortfrancois.github.io/samples/web-bluetooth/
https://beaufortfrancois.github.io/samples/web-bluetooth/write-descriptor.html
https://beaufortfrancois.github.io/samples/web-bluetooth/write-descriptor-async-await.html

R: @scheib

### DIFF
--- a/web-bluetooth/write-descriptor-async-await.html
+++ b/web-bluetooth/write-descriptor-async-await.html
@@ -1,0 +1,50 @@
+---
+feature_name: Web Bluetooth / Characteristic User Description (Async Await)
+chrome_version: 58
+check_min_version: true
+feature_id: 5264933985976320
+icon_url: icon.png
+index: index.html
+---
+
+{% include_relative _includes/intro.html %}
+
+<p>This sample illustrates the use of the Web Bluetooth API to change
+Characteristic User Description characteristic's descriptor of a service from a
+nearby Bluetooth Low Energy Device. You may want to try this demo with the BLE
+Peripheral Simulator App from the <a target="_blank"
+href="https://play.google.com/store/apps/details?id=io.github.webbluetoothcg.bletestperipheral">Google
+Play Store</a> and check out the <a
+href="write-descriptor.html">Characteristic User Description (Promises)</a> sample.</p>
+
+<p>
+  <input id="service" type="text" list="services" autofocus placeholder="Bluetooth Service">
+  <input id="characteristic" type="text" list="characteristics" placeholder="Bluetooth Characteristic">
+  <button id="readButton">Get Characteristic User Description</button>
+</p>
+
+{% include_relative _includes/datalist-services.html %}
+{% include_relative _includes/datalist-characteristics.html %}
+
+<p>
+  <input id="description" type="text" placeholder="Characteristic User Description">
+  <button id="writeButton" disabled>Set Characteristic User Description</button>
+</p>
+
+{% include output_helper.html %}
+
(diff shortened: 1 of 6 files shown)


web-bluetooth/write-descriptor-async-await.html
```diff
@@ -0,0 +1,50 @@
+---
+feature_name: Web Bluetooth / Characteristic User Description (Async Await)
```
scheib: update feature name to more clearly communicate that this is about writing to a descriptor.

web-bluetooth/write-descriptor-async-await.html
```diff
@@ -0,0 +1,50 @@
...
+check_min_version: true
+feature_id: 5264933985976320
+icon_url: icon.png
+index: index.html
+---
+
+{% include_relative _includes/intro.html %}
+
+<p>This sample illustrates the use of the Web Bluetooth API to change
+Characteristic User Description characteristic's descriptor of a service from a
```
scheib: ... to write to the descriptor 'Characteristic User Description' on a nearby Bluetooth ...

web-bluetooth/write-descriptor-async-await.js
```diff
@@ -0,0 +1,61 @@
...
+
+  let characteristicUuid = document.querySelector('#characteristic').value;
+  if (characteristicUuid.startsWith('0x')) {
+    characteristicUuid = parseInt(characteristicUuid);
+  }
+
+  try {
+    log('Requesting any Bluetooth Device...');
+    const device = await navigator.bluetooth.requestDevice({
+     // filters: [...] <- Prefer filters to save energy & show relevant devices.
```
scheib, line 17: I'm not sure we should acceptAllDevices in this sample. Depends on if devices advertise services that have this descriptor type. 

I do think we should suggest a few example services that have characteristics with this descriptor. Otherwise it may be unlikely for anyone to run this and have any success. Or, maybe we should really strongly encourage the use of BLE Sim app?

web-bluetooth/write-descriptor-async-await.html
beaufortfrancois: Done.

web-bluetooth/write-descriptor-async-await.html
beaufortfrancois: Done

web-bluetooth/write-descriptor-async-await.js
beaufortfrancois: From what I've experienced, it is easier to play with our samples when I use `acceptAllDevices` as the service is not advertised as you've noted.

My goal is to update our BLE Simulator App to add the ability to write/read this descriptor. See https://github.com/WebBluetoothCG/ble-test-peripheral-android/pull/76
I'm already suggesting it in the top paragraph.

beaufortfrancois:
@scheib 2/3 of your feedback is addressed.

scheib:
LGTM with text changes. I agree mostly with acceptAllDevices, I just didn't know if this sample would be easy for folks to get to work in any case without specifying a service & characteristic.

beaufortfrancois:
@g-ortuno Can you review https://github.com/WebBluetoothCG/ble-test-peripheral-android/pull/76 so that we have an android app people can use to play with GATT descriptors?

PR #482, opened 2017-02-04
Update README.md

entonbiba:
Thought I'd include a link to my sample here which makes the wavesurfer.js library work with the new Media Session API.
https://github.com/entonbiba/Media-Session-API-wavesurfer-js

### DIFF
--- a/media-session/README.md
+++ b/media-session/README.md
@@ -3,3 +3,7 @@ Media Session Sample
 See https://googlechrome.github.io/samples/media-session/index.html for a live demo.
 
 Learn more at https://www.chromestatus.com/feature/5639924124483584
+
+=== 
+<b>Other samples:</b><br/>
+wavesurfer.js with media session api - https://github.com/entonbiba/Media-Session-API-wavesurfer-js


entonbiba:
@googlebot I signed it!

beaufortfrancois:
Thanks for this @entonbiba!
Instead of the README file, I've added your cool sample at https://www.chromestatus.com/feature/5639924124483584 where all media session samples are gathered.

xxyzzzq:
@entonbiba @beaufortfrancois The demo doesn't work sadly.

* I see wavesurfer.js uses WebAudio to play an audio src, but Chrome on Android currently do not detect WebAudio playback state, so we consider the page as "not playing anything" and won't show a notification. A workaround is to create an audio element with volume 0 (don't use `muted` attribute), and keep the playback state consistent with wavesurfer.js.
* Also there's a misuse in the demo: currently "stop" action is not in the spec yet.

beaufortfrancois:
Thanks @xxyzzzq! I may re-add 
https://github.com/entonbiba/Media-Session-API-wavesurfer-js when it is working then.

entonbiba:
@beaufortfrancois yes just a quick sample, I'll fix it up once the API is fully supported by Chrome. It would be nice if the official standard also supports stop action. :)

PR #483, opened 2017-02-13
Add video media session sample

beaufortfrancois:
Following @mounirlamouri suggestion, here's a media session sample featuring video.

You can try this PR at:
http://beaufortfrancois.github.io/samples/media-session/video.html
http://beaufortfrancois.github.io/samples/media-session/audio.html
http://beaufortfrancois.github.io/samples/media-session/

R: @samdutton

### DIFF
--- a/media-session/audio.html
+++ b/media-session/audio.html
@@ -1,0 +1,40 @@
+---
+feature_name: Audio / Media Session
+chrome_version: 57
+feature_id: 5639924124483584
+check_min_version: true
+index: index.html
+---
+
+<h3>Background</h3>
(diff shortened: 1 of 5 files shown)


xxyzzzq:
FYI: You can add play/pause action handlers once https://crbug.com/689637 is fixed. Currently it's broken for videos as background videos cannot be resumed.

mounirlamouri:
lgtm 👍  As @xxyzzzq said, we would need to handle `play` and `pause` actions but right now, Chrome 57 has a bug. I'm surprised (and glad) to see that `previoustrack` and `nexttrack`.

samdutton:
LGTM 🙏

PR #485, opened 2017-02-24
Add play & pause Media Session action handlers

beaufortfrancois:
Now that https://codereview.chromium.org/2681863005 is fixed, let's add missing Media Session action handlers.

Merge when M57 hits stable on Chrome for Android.

TBR: @xxyzzzq @mounirlamouri


mounirlamouri:
I think this is ready to be merged :)

beaufortfrancois:
Thank you @mounirlamouri.
I'll push then!

PR #496, opened 2017-03-23
No longer desktop only and behind a flag.

mounirlamouri:


mounirlamouri:
@beaufortfrancois can you PTAL? :)

beaufortfrancois:
Thanks!

PR #502, opened 2017-03-28
[Presentation API] Adds a receiver page that accepts connections & shows fruits

markafoltz:
@beaufortfrancois

### DIFF
--- a/presentation-api/index.html
+++ b/presentation-api/index.html
@@ -1,7 +1,7 @@
 ---
-feature_name: Presentation API
-chrome_version: 51
-feature_id: 6676265876586496
+feature_name: Presentation API: Receiver API
+chrome_version: 58
+feature_id: 5414209298890752
 ---
 
 <h3>Background</h3>
@@ -10,7 +10,7 @@ feature_id: 6676265876586496
   This sample illustrates the use of
   <a href="https://w3c.github.io/presentation-api/">Presentation API</a>,
   which gives the ability to access external presentation-type displays and use
-  them for presenting web content. The <code>PresentationRequest</code> object
+  them for presenting Web content. The <code>PresentationRequest</code> object
   is associated with a request to initiate or reconnect to a presentation made
   by a controlling browsing context and it takes in a presentation request URL
   when constructed. A presentation can be started by calling the
@@ -36,28 +36,30 @@ feature_id: 6676265876586496
 {% include output_helper.html initial_output_content=initial_output_content %}
 
 {% capture js %}
-// App ID EF1A139F is registered in the Google Cast SDK Developer
-// Console and points to the following custom receiver:
-// https://googlechrome.github.io/samples/presentation-api/receiver/index.html
-var presentationUrl = 'https://google.com/cast/#__castAppId__=EF1A139F';
-var presentationRequest;
-var presentationConnection;
+const presentationUrls =
+  ['https://googlechrome.github.io/samples/presentation-api/receiver/index.html']
+let presentationRequest;
+let presentationConnection;
 
 function updatePresentationAvailability(displaysAvailable) {
   ChromeSamples.log('Presentation displays are ' +
       (displaysAvailable ? 'available' : 'unavailable') + '.');
 }
 
-document.querySelector('#createRequest').addEventListener('click', function() {
-  presentationRequest = new PresentationRequest(presentationUrl);
+document.querySelector('#createRequest').addEventListener('click', () => {
+  presentationRequest = new PresentationRequest(presentationUrls);
   document.querySelector('#createRequest').disabled = true;
   document.querySelector('#start').disabled = false;
   document.querySelector('#reconnect').disabled = false;
-  presentationRequest.getAvailability().then(function(availability) {
+  presentationRequest.getAvailability().then(availability => {
     updatePresentationAvailability(availability.value);
-    availability.onchange = function() {
+    availability.onchange = () => {
       updatePresentationAvailability(availability.value);
     };
+  }).catch(error => {
+    ChromeSamples.log('Presentation availability not supported, ' +
+                      error.name + ': ' + error.message);
+    updatePresentationAvailability(false);
   });
 });
 
@@ -65,36 +67,37 @@ function setConnection(connection) {
   presentationConnection = connection;
   document.querySelector('#close').disabled = false;
   document.querySelector('#terminate').disabled = false;
-  presentationConnection.onclose = function() {
+  presentationConnection.onclose = () => {
     ChromeSamples.log('Connection closed.');
   };
-  presentationConnection.onterminate = function() {
+  presentationConnection.onterminate = () => {
     ChromeSamples.log('Connection terminated.');
   };
 }
 
-document.querySelector('#start').addEventListener('click', function() {
-  presentationRequest.start().then(function(connection) {
+document.querySelector('#start').addEventListener('click', () => {
+  presentationRequest.start().then(connection => {
     ChromeSamples.log('Connected to presentation: ' + connection.id);
     setConnection(connection);
-  }).catch(function(error) {
+  }).catch(error => {
     ChromeSamples.log(error.name + ': ' + error.message);
   });
 });
 
-document.querySelector('#reconnect').addEventListener('click', function() {
-  var presentationId = document.querySelector('#presentationId').value.trim();
+document.querySelector('#reconnect').addEventListener('click', () => {
+  const presentationId = document.querySelector('#presentationId').value.trim();
   if (presentationId.length) {
-    presentationRequest.reconnect(presentationId).then(function(connection) {
+    presentationRequest.reconnect(presentationId).then(connection => {
       ChromeSamples.log('Reconnected to presentation: ' + connection.id);
       setConnection(connection);
-    }).catch(function(error) {
-      ChromeSamples.log(error.name + ': ' + error.message);
+    }).catch(error => {
+      ChromeSamples.log('Presentation.reconnect() error, ' + error.name + ': ' +
+                        error.message);
     });
   }
 });
 
-document.querySelector('#close').addEventListener('click', function() {
+document.querySelector('#close').addEventListener('click', () => {
   if (presentationConnection) {
     presentationConnection.close();
     document.querySelector('#close').disabled = true;
@@ -102,7 +105,7 @@ document.querySelector('#close').addEventListener('click', function() {
   }
 });
 
-document.querySelector('#terminate').addEventListener('click', function() {
+document.querySelector('#terminate').addEventListener('click', () => {
   if (presentationConnection) {
     presentationConnection.terminate();
     document.querySelector('#close').disabled = true;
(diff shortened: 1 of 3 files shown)


markafoltz:
@zhaobin2016

presentation-api/index.html
```diff
@@ -38,6 +42,7 @@
 {% capture js %}
 const presentationUrls =
   ['https://googlechrome.github.io/samples/presentation-api/receiver/index.html'];
```
beaufortfrancois: I would recommend a relative URL so that it works when testing:
```js
const presentationUrls = ['receiver/index.html'];
```

presentation-api/receiver/index.html
```diff
@@ -17,15 +17,81 @@
...
+          'banana':      '\u{1F34C}',
+          'pineapple':   '\u{1F34D}',
+          'green apple': '\u{1F35F}',
+          'apple':       '\u{1F34E}',
+          'pear':        '\u{1F350}',
+          'peach':       '\u{1F351}',
+          'cherries':    '\u{1F352}',
+          'strawberry':  '\u{1F353}'
+        };
+        const addConnection = (connection) => {
```
beaufortfrancois: Do you mind using `function addConnection(connection)` instead everywhere?

presentation-api/receiver/index.html
```diff
@@ -17,15 +17,81 @@
...
+        const maybeSetFruit = (message) => {
+          const fruit = Object.entries(fruitEmoji).find(
+            ([name, character]) => {return message.indexOf(name) > -1}
+          );
+          if (fruit) {
+            document.getElementById("main").textContent = fruit[1];
+          }
+        };
+
+        navigator.presentation.receiver &&
```
beaufortfrancois: How about?
```js
if (navigator.presentation.receiver) {
  navigator.presentation.receiver.connectionList.then(list => {
    list.connections.map(connection => {
      addConnection(connection);
    });
    list.onconnectionavailable = (evt) => {
      addConnection(evt.connection);
    };
  });
}

File: presentation-api/receiver/index.html [hunk 4]
```diff
@@ -17,15 +17,81 @@
...
+          };
+
+          connection.onclose = (closeEvent) => {
+            addMessage('Connection #' + connection.connectionId + ' closed, reason = ' +
+                       closeEvent.reason + ', message = ' + closeEvent.message); 
+          };
+        };
+
+        const addMessage = (content, language) => {
+          const listElt = document.getElementById("message-list");
```
beaufortfrancois: `s/getElementById/querySelector('#/g`

presentation-api/index.html
```diff
@@ -97,6 +104,15 @@
   }
 });
 
+document.querySelector('#send').addEventListener('click', () => {
+  if (presentationConnection && presentationConnection.state == 'connected') {
+    const message = document.querySelector('#message').value.trim();
+    const lang = document.body.lang || 'en-US';
+    ChromeSamples.log('Sending ' + message + ' to ' + presentationConnection.id);
+    presentationConnection.send(JSON.stringify({string: message, lang: lang}));
```
beaufortfrancois: Do we need to send `lang` as well? If not, I'd suggest simply sending `message`.

beaufortfrancois: If we rename `string` to `message`, we could simply do:

```js
presentationConnection.send(JSON.stringify({message}));
```

presentation-api/receiver/index.html
```diff
@@ -17,15 +17,81 @@
...
+            navigator.presentation.receiver.connectionList.then(list => {
+          list.connections.map(connection => {
+            addConnection(connection);
+          });
+          list.onconnectionavailable = (evt) => {
+            addConnection(evt.connection);
+          };
+        });
+      })();
+    <script>
```
beaufortfrancois: `s/<script>/</script>`

presentation-api/receiver/index.html
```diff
@@ -17,15 +17,81 @@
         left: 50%;
         transform: translate(-50%, -50%);
         text-align: center;
-        font-size: 75px;
+        font-size: 150px;
```
beaufortfrancois: `150px` is too big (e.g. it doesn't show up). Use `128px` instead for unicode.

presentation-api/index.html
```diff
@@ -29,6 +29,10 @@
   <button id="reconnect" disabled>presentationRequest.reconnect(presentationId)</button>
 </p>
 <p>
+  <input id="message" type="text" placeholder="Try a fruit!">
```
beaufortfrancois: Please use a datalist that contains all fruits we handle so that it is easier for user to play with:
```html
<input id="message" type="text" placeholder="Try a fruit!" list="fruits">

<datalist id="fruits">
  <option>grapes</option>
  ...
</datalist>

File: presentation-api/index.html [hunk 9]
```diff
@@ -29,6 +29,10 @@
   <button id="reconnect" disabled>presentationRequest.reconnect(presentationId)</button>
 </p>
 <p>
```
beaufortfrancois, line 46: May we move this paragraph (fruit + button) above "reconnect"?

presentation-api/receiver/index.html
```diff
@@ -17,15 +17,81 @@
...
+        const addMessage = (content, language) => {
+          const listElt = document.getElementById("message-list");
+          const listItem = document.createElement("LI");
+          if (language) listItem.lang = language;
+          listItem.textContent = content;
+          listElt.appendChild(listItem);
+        };
+
+        const maybeSetFruit = (message) => {
+          const fruit = Object.entries(fruitEmoji).find(
```
beaufortfrancois: You may want to simplify this code with:

```js
  if (message in fruitEmoji) {
    document.querySelector('#main').textContent = fruitEmoji[message]; 
  }
```

beaufortfrancois (commented):
See my 2 cents.

presentation-api/index.html
beaufortfrancois: If so, we may want to log `presentationConnection.url` as well.

presentation-api/receiver/index.html
```diff
@@ -17,15 +17,81 @@
         left: 50%;
         transform: translate(-50%, -50%);
         text-align: center;
-        font-size: 75px;
+        font-size: 150px;
       }
     </style>
+    <script type="text/javascript">
```
beaufortfrancois: We should add this as a separate script we could link from the sample so that developers understand how works communication.

presentation-api/index.html
markafoltz: Done.

presentation-api/index.html
markafoltz: Done

presentation-api/index.html
markafoltz: Done

presentation-api/index.html
markafoltz: It's recommended to send the language of human readable text so the remote page can render it with the appropriate lang attribute.

markafoltz: Not sure I follow.  The receiver page expects the message content to be in the `string` property.

presentation-api/receiver/index.html
markafoltz: 150px worked for me, but done anyway.

presentation-api/receiver/index.html
markafoltz: Done.

presentation-api/receiver/index.html
markafoltz: Done

presentation-api/receiver/index.html
markafoltz: Done.

presentation-api/receiver/index.html
markafoltz: That wouldn't provide case insensitive matching.

presentation-api/receiver/index.html
markafoltz: OK.  Done.

presentation-api/receiver/index.html
markafoltz: No longer needed.

markafoltz:
Okay to merge @beaufortfrancois ?

presentation-api/index.html
beaufortfrancois: @mfoltzgoogle `list=fruits` is still missing to make this work.

presentation-api/index.html
beaufortfrancois: Noted.
I can see this at https://www.w3.org/TR/presentation-api/#passing-locale-information-with-a-message but I don't get why the remote page needs it. if the appropriate charset is defined, we should be good no? (e.g. ` <meta charset="UTF-8">`)

beaufortfrancois: If we were to rename `string` key to `message` in the receiver page, we could do something like

```js
presentationConnection.send(JSON.stringify({message, lang}));
```

It is just a suggestion for readability.

presentation-api/receiver/index.html
beaufortfrancois: It looks like `receiver.js` has not been uploaded to this PR.

presentation-api/receiver/index.html
beaufortfrancois: You're right.
Let' do then:


```js
  if (message in fruitEmoji) {
    document.querySelector('#main').textContent = fruitEmoji[message.toLowerCase()]; 
  }
```

beaufortfrancois (commented):
@mfoltzgoogle See my feedback

presentation-api/index.html
markafoltz: The issue is that the lang attribute affects font rendering in CJK.  See https://github.com/w3c/presentation-api/issues/218

markafoltz: Ah, I see.  Done.

presentation-api/receiver/index.html
markafoltz: Sorry, including receiver.js now.

presentation-api/receiver/index.html
markafoltz: Done.

presentation-api/index.html
markafoltz: I wanted to also allow arbitrary messages...  Added a second input then.

markafoltz:
PTAL @beaufortfrancois

presentation-api/index.html
beaufortfrancois: `"list=fruits"` does not force user to pick a fruit, it is a simple suggestion list.
Therefore, I'd suggest we remove "sendMessage" there.

presentation-api/receiver/index.html
beaufortfrancois: Nit: do you mind removing trailing white spaces in that file?

beaufortfrancois:
LGTM with nits

markafoltz:
@beaufortfrancois Can this be merged now?  I don't have write access. Thx

beaufortfrancois:
Just merged! It is now live at https://googlechrome.github.io/samples/presentation-api/

beaufortfrancois:
I've added this sample to https://www.chromestatus.com/feature/5414209298890752 as well

beaufortfrancois:
And removed trailing whitespaces at https://github.com/GoogleChrome/samples/pull/516

PR #504, opened 2017-03-30
Add the necessary connection logic

avayvod:
As prescribed when using the Cast Receiver SDK here:  https://developers.google.com/cast/docs/custom_receiver#initialization--readiness


beaufortfrancois:
👍

PR #508, opened 2017-04-06
[Web Payments] Rename buyButton because Safari doesn't like it

madmath:
"can't create duplicate variable that shadows a global property"

const buyButton = document.getElementById('buyButton')

renaming the global to 'payButton' fixes that.


rsolomakhin (approved):
LGTM

madmath:
@beaufortfrancois to merge, thanks

PR #509, opened 2017-04-20
2017 links update

dandv:
Also, https://developers.google.com/web/updates/2014/11/Support-for-installable-web-apps-with-webapp-manifest-in-chrome-38-for-Android should be marked as deprecated and point to https://developers.google.com/web/fundamentals/engage-and-retain/app-install-banners/


PaulKinlan:
Sorry for delay, thanks

PR #513, opened 2017-05-16
Remove hover on-demand in media-hover-pointer demo.

darktears:
The value was removed from the spec (see https://drafts.csswg.org/mediaqueries-4/#descdef-media-hover).

The support was also removed from Chrome https://bugs.chromium.org/p/chromium/issues/detail?id=654861.

Ecosystem is aligning see https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9308676/

Firefox and Safari never shipped this property.

### DIFF
--- a/media-hover-pointer/index.html
+++ b/media-hover-pointer/index.html
@@ -140,12 +140,6 @@ limitations under the License.
         text-decoration: none;
       }
     }
-    @media (any-hover: on-demand) {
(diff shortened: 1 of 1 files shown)


darktears:
@RByers I need someone with super power :).

RByers:
Change LGTM, thanks!  But I don't have push access.  @samthor @jeffposnick can one of you please merge?

PR #514, opened 2017-05-18
Link update to current specification

hexalys:
Link update to current specification w/o the now removed 'on-demand' to avoid any confusion.


PaulKinlan:
Sorry for delay, thanks

PR #517, opened 2017-05-29
Add VP9 codec string sample

beaufortfrancois:
Hello @wolenetz,
This sample describes the upcoming new VP9 codec string and Profile 2 support in Chrome 60.

What do you think?

screenshot 2017-05-29 at 1 04 20 pm - display 1

### DIFF
--- a/media/vp9-codec-string.js
+++ b/media/vp9-codec-string.js
@@ -1,0 +1,25 @@
+if (MediaSource.isTypeSupported('video/webm; codecs="vp9"')) {
+  log('> VP9, Profile 0, level 1 (later fields defaulted)');
+}
+
+if (MediaSource.isTypeSupported('video/webm; codecs="vp09.00.10.08"')) {
+  log('> VP9, Profile 0, level 1, bit depth 8 (later fields defaulted)');
+}
+
+if (MediaSource.isTypeSupported('video/webm; codecs="vp09.01.20.08.01"')) {
+  log('> VP9, Profile 0, level 1, bit depth 8, 4:2:0 chroma subsampling ' +
+      '(later fields defaulted)');
+}
+
+if (MediaSource.isTypeSupported('video/webm; codecs="vp09.01.20.08.01.01.01.01.00"')) {
+  log('> VP9, Profile 1, level 2, bit depth 8, ' +
+      '4:2:0 chroma subsampling, REC709 color/transfer/matrix, ' +
+      'luma/chroma encoded in the "legal" range.');
+}
+
+if (MediaSource.isTypeSupported('video/webm; codecs="vp09.02.10.10.01.09.16.09.01"')) {
+  log('> VP9, Profile 2, level 1, 10-bit YUV content, ' +
+      '4:2:0 chroma subsampling, ITU-R BT.2020 primaries, ' + 
+      'ST 2084 EOTF, ITU-R BT.2020 non-constant luminance color matrix, ' +
+      'full-range chroma/luma encoding.');
+}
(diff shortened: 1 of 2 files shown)


media/vp9-codec-string.js
```diff
@@ -0,0 +1,25 @@
+if (MediaSource.isTypeSupported('video/webm; codecs="vp9"')) {
+  log('> VP9, Profile 0, level 1 (later fields defaulted)');
```
chcunningham: I've always thought of the the old string is completely ambiguous for profile/level. I've not been around long enough to say whether it was originally intended to just imply profile 0, but in practice its used (regrettably) for lots of other profiles/levels.

media/vp9-codec-string.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+if (MediaSource.isTypeSupported('video/webm; codecs="vp09.00.10.08"')) {
+  log('> VP9, Profile 0, level 1, bit depth 8 (later fields defaulted)');
+}
+
+if (MediaSource.isTypeSupported('video/webm; codecs="vp09.01.20.08.01"')) {
+  log('> VP9, Profile 0, level 1, bit depth 8, 4:2:0 chroma subsampling ' +
```
chcunningham: Log doesn't match the string. String shows profile 1, level 2. Sorry, I made this typo in the web platform tests.

chcunningham: nit: 4:2:0 chroma subsampling is correct, but could apply to eitehr values of 0 or 1. I realize my test documentation is equally ambiguous. You might help clarify that this is value 4:2:0 variety that includes "colocated with luma (0,0)" ... or similar.

chcunningham (commented):
Thanks for making these samples!

media/vp9-codec-string.js
beaufortfrancois: @chcunningham What should it say then?
```js
log('Some profile/level of VP9');
```

media/vp9-codec-string.js
beaufortfrancois: Thank you!

beaufortfrancois:
@chcunningham I've addressed your feedback at https://github.com/GoogleChrome/samples/pull/517/commits/9d0d00eb87b5b646f3b96622bb21296af878a8e2. 
Let me know it that looks good to you.

beaufortfrancois:
By the way, I've tried with the `--enable-new-vp9-codec-string ` switch in Chrome OS and I get:

```
Your browser supports:
> Some VP9 profile
> VP9, Profile 0, level 1, bit depth 8 (later fields defaulted)
> VP9, Profile 1, level 2, bit depth 8, 4:2:0 chroma subsampling colocated with (0,0) luma, (later fields defaulted)
> VP9, Profile 1, level 2, bit depth 8, 4:2:0 chroma subsampling colocated with (0,0) luma, REC709 color/transfer/matrix, luma/chroma encoded in the "legal" range.
```

Profile 2 is missing. Is that intended?

chcunningham:
LGTM.

Profile 2 is supported, but your getting "false" because the specified eotf and color primaries are not supported unless chrome is running in "HDR mode". This is still under development, but you can enable HDR mode by running with --enable-hdr.

You can see which color primaries/transfer/matrices require HDR mode (implicitly "color management") here:
https://cs.chromium.org/chromium/src/media/base/decode_capabilities.cc?q=color_management&l=15

If you're hoping for profile 2 to be more reliably supported you can choose a different eotf and color primaries.

beaufortfrancois:
@chcunningham You're right. It works great with `--enable-hdr`

wolenetz:
(@ https://github.com/GoogleChrome/samples/pull/517#issue-231993643, @chcunningham is the correct reviewer).
Aside - I also think just "vp9" was ambiguous w.r.t. profile/level/etc.

PR #527, opened 2017-07-24
Add choose file button to error-message sample

beaufortfrancois:
I found myself using that often these days and thought it would benefit more users. So let's add a "Choose File" button to the official MediaError.message Sample.

Try it live at https://beaufortfrancois.github.io/samples/media/error-message.html

What do you think @wolenetz?

### DIFF
--- a/media/error-message.js
+++ b/media/error-message.js
@@ -24,6 +24,6 @@ chooseFileButton.addEventListener('change', function() {
 });
 
 video.addEventListener('canplay', function() {
-  ChromeSamples.log('Video can play!');
+  ChromeSamples.log('"canplay" event received');
 });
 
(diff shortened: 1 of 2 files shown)


media/error-message.js
```diff
@@ -17,3 +17,13 @@ nonExistentFileButton.addEventListener('click', function() {
...
   video.src = 'https://storage.googleapis.com/media-error/no_streams.webm';
 });
+
+chooseFileButton.addEventListener('change', function() {
+  const file = chooseFileButton.files[0];
+  video.src = URL.createObjectURL(file);
+});
+
+video.addEventListener('canplay', function() {
+  ChromeSamples.log('Video can play!');
```
wolenetz: Nit: I suggest something more like "'canplay' event received" instead, because:
1) it might not be a video
2) "Video can play!" might overstate the ability of the pipeline to actually decode and render the media.

Otherwise LGTM.

beaufortfrancois:
@wolenetz I've addressed your feedback. Thanks!

beaufortfrancois:
It is now live at https://googlechrome.github.io/samples/media/error-message.html

PR #536, opened 2017-08-17
Add MediaSource.setLiveSeekableRange Sample

beaufortfrancois:
Hello @wolenetz,

What do you think of this sample?
There will be a post for M62 like https://developers.google.com/web/updates/2017/07/chrome-61-media-updates to explain what this is later.

Preview at https://beaufortfrancois.github.io/samples/media/live-seekable-range.html

### DIFF
--- a/media/live-seekable-range.html
+++ b/media/live-seekable-range.html
@@ -41,29 +41,65 @@ result in a single seekable range which fits both, when media duration is
 
 <script>
   document.querySelector('#setInfiniteDurationButton').addEventListener('click', function() {
-    onSetInfiniteDurationButtonClick();
+    try {
+      onSetInfiniteDurationButtonClick();
+    } catch(error) {
+      logError(error);
+    }
   });
   document.querySelector('#fetchAndAppend3To6MediaButton').addEventListener('click', function() {
-    onFetchAndAppend3To6MediaButtonClick();
+    try {
+      onFetchAndAppend3To6MediaButtonClick();
+    } catch(error) {
+      logError(error);
+    }
   });
   document.querySelector('#set1Button').addEventListener('click', function(event) {
-    onSetLiveSeekableRangeButtonClick(event);
+    try {
+      onSetLiveSeekableRangeButtonClick(event);
+    } catch(error) {
+      logError(error);
+    }
   });
   document.querySelector('#set2Button').addEventListener('click', function(event) {
-    onSetLiveSeekableRangeButtonClick(event);
+    try {
+      onSetLiveSeekableRangeButtonClick(event);
+    } catch(error) {
+      logError(error);
+    }
   });
   document.querySelector('#clearButton').addEventListener('click', function() {
-    onClearLiveSeekableRangeButtonClick();
+    try {
+      onClearLiveSeekableRangeButtonClick();
+    } catch(error) {
+      logError(error);
+    }
   });
   document.querySelector('#removeButton').addEventListener('click', function() {
-    onRemoveMediaSegmentButtonClick();
+    try {
+      onRemoveMediaSegmentButtonClick();
+    } catch(error) {
+      logError(error);
+    }
   });
   document.querySelector('#setFiniteDurationButton').addEventListener('click', function() {
-    onSetFiniteDurationButtonClick();
+    try {
+      onSetFiniteDurationButtonClick();
+    } catch(error) {
+      logError(error);
+    }
   });
   document.querySelector('#endOfStreamButton').addEventListener('click', function() {
-    onEndOfStreamButtonClick();
+    try {
+      onEndOfStreamButtonClick();
+    } catch(error) {
+      logError(error);
+    }
   });
+
+  function logError(error) {
+    log(`> Error: ${error.message}\n`);
+  }
 
   log = ChromeSamples.log;
 </script>
--- a/media/live-seekable-range.html
+++ b/media/live-seekable-range.html
@@ -11,26 +11,58 @@ This API lets MediaSource apps more effectively customize the
 <code>HTMLMediaElement.seekable</code> range logic by providing (or removing) a
 single seekable range that is union'ed with the current buffered ranges to
 result in a single seekable range which fits both, when media duration is
-<code>+Infinite</code>.
+<code>+Infinity</code>.
 </p>
 
-<button id="setButton">mediaSource.setLiveSeekableRange(40, 42)</button>
+<style>
+  video {
+    background: #1e2327;
+    border-radius: 4px;
+    width: 100%;
+  }
+  button {
+    margin: 4px 0;
+  }
+</style>
+
+<video controls></video>
+<button id="setInfiniteDurationButton">mediaSource.duration = +Infinity</button>
+<button id="fetchAndAppend3To6MediaButton">fetch & append 3 to 6 seconds media segment</button>
+<button id="set1Button">mediaSource.setLiveSeekableRange(40, 42)</button>
+<button id="set2Button">mediaSource.setLiveSeekableRange(1, 4)</button>
 <button id="clearButton">mediaSource.clearLiveSeekableRange()</button>
 <button id="removeButton">sourceBuffer.remove(3, 6)</button>
+<button id="setFiniteDurationButton">mediaSource.duration = 6</button>
+<button id="endOfStreamButton">mediaSource.endOfStream()</button>
 
 {% include output_helper.html initial_output_content=initial_output_content %}
 
 {% include js_snippet.html filename='live-seekable-range.js' %}
 
 <script>
-  document.querySelector('#setButton').addEventListener('click', function() {
-    onSetLiveSeekableRangeButtonClick();
+  document.querySelector('#setInfiniteDurationButton').addEventListener('click', function() {
+    onSetInfiniteDurationButtonClick();
+  });
+  document.querySelector('#fetchAndAppend3To6MediaButton').addEventListener('click', function() {
+    onFetchAndAppend3To6MediaButtonClick();
+  });
+  document.querySelector('#set1Button').addEventListener('click', function(event) {
+    onSetLiveSeekableRangeButtonClick(event);
+  });
+  document.querySelector('#set2Button').addEventListener('click', function(event) {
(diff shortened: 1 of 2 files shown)


media/live-seekable-range.html
```diff
@@ -0,0 +1,37 @@
...
+check_min_version: true
+---
+
+<h3>Background</h3>
+<p>
+This API lets MediaSource apps more effectively customize the
+<code>HTMLMediaElement.seekable</code> range logic by providing (or removing) a
+single seekable range that is union'ed with the current buffered ranges to
+result in a single seekable range which fits both, when media duration is
+<code>+Infinite</code>.
```
wolenetz: +Infinity

media/live-seekable-range.js
```diff
@@ -0,0 +1,75 @@
...
+  log('User clicked "sourceBuffer.remove(3, 6)" button');
+  mediaSource.sourceBuffers[0].remove(3 /* start */, 6 /* end*/);
+  mediaSource.sourceBuffers[0].addEventListener('updateend', function() {
+    logTimeRanges();
+  }, { once: true });
+}
+
+/* Util */
+
+function logTimeRanges() {
```
wolenetz: Please also log video.duration and mediaSource.duration.

media/live-seekable-range.html
```diff
@@ -0,0 +1,37 @@
+---
```
wolenetz, line 1: Strange - on Linux, Google Chrome 60.0.3112.101 (Official Build) (64-bit), it looks like this API is supported. Maybe Blink's MediaSourceExperimental RuntimeEnabledFlag is enabled for Googlers.
Regardless, I observed eventual "Script error. (Your browser may not support this feature.)" -- Other than clicking a bunch of the buttons and waiting, and waiting some more, then switching tabs/windows and coming back, I'm not sure how to repro that error. Note that initially that same row in the log shows just "Warning! Keep in mind this sample has been tested with Chrome 62."

media/live-seekable-range.html
```diff
@@ -0,0 +1,37 @@
...
+  document.querySelector('#setButton').addEventListener('click', function() {
+    onSetLiveSeekableRangeButtonClick();
+  });
+  document.querySelector('#clearButton').addEventListener('click', function() {
+    onClearLiveSeekableRangeButtonClick();
+  });
+  document.querySelector('#removeButton').addEventListener('click', function() {
+    onRemoveMediaSegmentButtonClick();
+  });
+
```
wolenetz: Three more buttons please:
1) setLiveSeekableRange(1,4)  --> result, if there's still buffered [3,6) and +Infinity duration, should be seekable [1,6)  --> This shows the front part of the "union" operation is working. (And if nothing buffered, should be just [1,6) if +Infinity duration).
2) setInfiniteDuration() --> mediaSource.duration = +Infinity; and then log result..
3) setFiniteDuration() --> mediaSource.duration = some finite value >= highest buffered end time. Maybe 6?

We might also want to add endOfStream() and fetchAndAppend3To6Media() to show the interesting behavior resulting from those calls.

wolenetz:
@mounirlamouri please take a look, too. This API impacts HTMLMediaElement.seekable, which native controls rely upon.

media/live-seekable-range.html
beaufortfrancois: Done.

media/live-seekable-range.js
beaufortfrancois: I've added `mediaSource.duration` but not `video.duration` as it is always the same as `mediaSource.duration`.

However, I've noticed something strange you may help me understand...
In `updateend` event, `video.duration` is `NaN` when I append init segment. If I wrap the call of `video.duration` in `requestAnimationFrame`, it is now equals to `mediaSource.duration`.
According to the [spec](https://www.w3.org/TR/media-source/#duration-change-algorithm), they should be set at the same time.

media/live-seekable-range.html
beaufortfrancois: It works as expected as I've added `check_min_version = true` at https://github.com/GoogleChrome/samples/pull/536/files#diff-14e684df0f667174c42ca3d4b20f7a15R5
It is just a way for user to know that this sample has been tested with Chrome 62 (which I did)

media/live-seekable-range.html
beaufortfrancois: Done

beaufortfrancois:
For info, https://bugs.chromium.org/p/chromium/issues/detail?id=756785 and https://bugs.chromium.org/p/chromium/issues/detail?id=756800 have been filed while playing with this sample.

media/live-seekable-range.html
beaufortfrancois: Regarding the `Script error.`, I can't reproduce with same config ;(

media/live-seekable-range.js
```diff
@@ -1,68 +1,88 @@
...
-    // Append video init segment.
-    sourceBuffer.appendBuffer(data);
-    sourceBuffer.addEventListener('updateend', function() {
-      logTimeRanges();
-
-      log('Fetching video segment that starts at 3 seconds...');
-      fetch(videoUrl, { headers: { range: 'bytes=567140-1196488' } })
-      .then(response => response.arrayBuffer())
-      .then(data => {
+  fetchAndAppendMediaSegment('bytes=0-299');
```
wolenetz: nit: to not confuse folks with "Media" segment for what is technically the bytestream segment corresponding to an MSE "Initialization Segment", perhaps rename this method to "fetchAndAppendSegment"?

beaufortfrancois: Fixed in https://github.com/GoogleChrome/samples/pull/536/commits/910dd1c4ab8d605bca8ab8c147cfb58b0ed7b126

media/live-seekable-range.js
```diff
@@ -1,68 +1,88 @@
...
-function onSetLiveSeekableRangeButtonClick() {
-  log('User clicked "mediaSource.setLiveSeekableRange(40, 42)" button');
-  mediaSource.setLiveSeekableRange(40 /* start */, 42 /* end */);
-  logTimeRanges();
+function onSetLiveSeekableRangeButtonClick(event) {
+  const re = /mediaSource\.setLiveSeekableRange\((\d+), (\d+)\)/;
+  const start = Number(re.exec(event.target.textContent)[1]);
+  const end = Number(re.exec(event.target.textContent)[2]);
+  log(`User clicked "mediaSource.setLiveSeekableRange(${start}, ${end})" button`);
+  mediaSource.setLiveSeekableRange(start, end);
```
wolenetz, line 32: Many of these operations depend on the mediaSource being in the 'open' state.
So, if a sequence contains endOfStream() without a subsequent fetchAndAppend or remove, then exception will occur, like "Uncaught InvalidStateError: Failed to set the 'duration' property on 'MediaSource': The MediaSource's readyState is not 'open'. (Your browser may not support this feature.)" for many of the other buttons.

Suggestion is to at least move that exception message to be in-line in the log with the sequence of operations.

Second suggestion is to catch the exception in the button click handler functions, so that they can still logMediaInfo after (and so they perhaps can give a better understanding of what's going on to the user).

wolenetz: Also, if 'updating' is true, similar exception can occur. This may happen if there is a slow network connection servicing the fetches, or if the user rapidly clicks buttons, and possibly in other cases. Suggestion: also catch and explain that kind of exception, or first check 'updating' state and explain why it's not a good idea to continue attempting the operation, or maybe even disable all buttons while a button's current processing is running still.

wolenetz (commented):
Looking pretty good. Thanks for putting this together!

media/live-seekable-range.html
wolenetz: I'm not sure where that script error was coming from.
Also, I confirmed I had set my M60's chrome://flags to enable experimental web platform features (which is why set/clearLiveSeekableRange is working for me). I'm sorry for that bit of review noise :/

media/live-seekable-range.js
wolenetz: Please file a bug around this. I know why it's happening. I already fixed a similar endOfStream duration change race recently in M62.

media/live-seekable-range.js
```diff
@@ -0,0 +1,95 @@
...
+video.src = URL.createObjectURL(mediaSource);
+  
+mediaSource.addEventListener('sourceopen', function() {
+  URL.revokeObjectURL(video.src);
+
+  mediaSource.addSourceBuffer('video/webm; codecs="vp9"');
+
+  log('Fetching video init segment...');
+  fetchAndAppendSegment('bytes=0-299');
+});
```
wolenetz: Make this handler only happen Once. Otherwise, error occurs later if endOfStream() followed by anything that re-opens the mediaSource (like append or remove) is done, because a re-opening sends a 'sourceopen' event again.

Example error:
Uncaught QuotaExceededError: Failed to execute 'addSourceBuffer' on 'MediaSource': This MediaSource has reached the limit of SourceBuffer objects it can handle. No additional SourceBuffer objects may be added. (Your browser may not support this feature.)

beaufortfrancois: Done.

media/live-seekable-range.js
beaufortfrancois: I've wrapped all button clicks events in try/catch to log errors so that user understand what is happening properly. Thanks for catching this!

beaufortfrancois:
@wolenetz We should be good now.
@mounirlamouri WDYT?

beaufortfrancois:
@wolenetz It that still looks good to you, I'll merge.

wolenetz:
LGTM -- Thanks for putting this together, @beaufortfrancois !!

Note related native controls work (at least https://bugs.chromium.org/p/chromium/issues/detail?id=756785 and https://bugs.chromium.org/p/chromium/issues/detail?id=756800) are not fixed yet in M62. (They're assigned to @mounirlamouri).

@mounirlamouri This sample should be quite helpful in exploring how native controls respond to buffered ranges, currentTime changes, seekable ranges, and duration changes.

beaufortfrancois:
Yeah! The official sample is now available at https://googlechrome.github.io/samples/media/live-seekable-range.html

PR #547, opened 2017-10-16
typo needed to compare to 'undefined' not undefined

arush:


arush:
there's a typo in your comments. It happens to be a highly ranking page on google and i copy pasted the incorrect code only to find out it had a typo

default-parameters-es6/demo.js
```diff
@@ -4,7 +4,7 @@
 function greetES5(message, to) {
   // Test parameter values in the function body, specifying a default
   to = to || 'DOM';
-  // OR if (typeof to === undefined) { to = 'DOM'; }
+  // OR if (typeof to === 'undefined') { to = 'DOM'; }
```
praveeno, line 7: i think `undefined` is right, because comment says, if `to` is undefined then `to = 'DOM'`. means if `to` is falsy value `undefined` then  assign `'dom'` into `to`.

praveeno (changes requested):

PR #553, opened 2017-11-09
Fixed TypeError caused by invoking generator function with 'new' keyword

nakhodkin:
Since generator functions returns iterator objects, they can't be as a constructor functions.
Thats why, they shouldn't be called with a `new` keyword.


nakhodkin:
I signed it!

PaulKinlan:
Sorry for the delay.

PR #555, opened 2017-11-14
Add Media Capabilities Decoding Sample

beaufortfrancois:
This sample will go with the upcoming "Audio/Video updates in Chrome 64" /web article.

Preview at https://beaufortfrancois.github.io/samples/media-capabilities/decoding

R: @mounirlamouri

### DIFF
--- a/media-capabilities/decoding.html
+++ b/media-capabilities/decoding.html
@@ -1,0 +1,25 @@
+---
+feature_name: Media Capabilities Decoding
+chrome_version: 64
+feature_id: 5869632707624960
+check_min_version: true
+---
+
+<h3>Background</h3>
+<p>
+The <a href="https://wicg.github.io/media-capabilities/">Media Capabilities
+ API</a> allows websites to get more information about the decoding
+abilities of the device/system/browser. This enables web developers to make
+optimal decisions when selecting media streams for the user.
+</p>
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
Note: I still need to add an Origin Trial token.

media-capabilities/decoding.html
```diff
@@ -0,0 +1,25 @@
...
+<h3>Background</h3>
+<p>
+The <a href="https://wicg.github.io/media-capabilities/">Media Capabilities
+ API</a> allows websites to get more information about the decoding
+abilities of the device/system/browser. This enables web developers to make
+optimal decisions when selecting media streams for the user.
+</p>
+<p>Note that the <code>result</code> will always be supported if the
+<code>mediaConfiguration</code> playback stats have not been recorded yet by
+the browser.</p>
```
mounirlamouri: Maybe say something like "Note that Chrome will always say that a supported configuration is smooth and powerEfficient until stats on this device have been recorded" or something like that. The main points are:
- Chrome, not "browsers"
- We say supported because it is, we pretend that it is smooth and efficient and that's what we don't really know. We know in advance if something is actually supported.

mounirlamouri (approved):
lgtm with comment applied

Issue #548, opened 2017-10-20
Add more referrer policies?


domfarolino:
I noticed that in https://github.com/GoogleChrome/samples/blob/gh-pages/fetch-api/fetch-referrer-policy.js we only have 5 referrer policies that we're testing. Is there a reason we don't have all of them? Perhaps the demo was created before all of the policies were finalized? I'd be happy to submit a PR.

Reference: https://w3c.github.io/webappsec-referrer-policy/#referrer-policies

domfarolino:
Bumping this thread. @jeffposnick thoughts?

jeffposnick:
My recollection is that those were the supported ones when I put together the demo. Feel free to PR in additional ones, and CC: me on the review!

PR #565, opened 2018-01-10
Add newer referrer policies

domfarolino:
Fixes https://github.com/GoogleChrome/samples/issues/548 by adding some missing referrer policies along with their descriptions. /cc @jeffposnick


jeffposnick:
Thanks!

PR #569, opened 2018-03-05
Add webassembly demos

surma:
These are demos for an upcoming Wasm article which I think can double-function as a sample. Please take a look if this is correct.

cc @jakearchibald @mathiasbynens


webassembly/README.md
```diff
@@ -0,0 +1,5 @@
+WebAssembly sample
+===
+See https://googlechrome.github.io/samples/webassembly/index.html for a live demo.
+
+Learn more at https://developers.google.com/web/updates/emscripting-a-c-library
```
beaufortfrancois: If you think* it will happen in march, I think it's safe to set the appropriate URL: https://developers.google.com/web/updates/2018/03/emscripting-a-c-library

surma: Good call

webassembly/README.md
```diff
@@ -0,0 +1,5 @@
+WebAssembly sample
+===
+See https://googlechrome.github.io/samples/webassembly/index.html for a live demo.
```
mathiasbynens, line 3: Could just be `https://googlechrome.github.io/samples/webassembly/` which looks nicer. For some reason we append `index.html` for other samples too…

mathiasbynens (approved):

PR #573, opened 2018-03-20
Add HDCP Policy Check sample

beaufortfrancois:
As HDCP Policy Check API is coming soon, here's the official sample that will be linked from the article.
Note that I'll merge only when upcoming article is live.

R: @xhwang-chromium

Preview URL: https://beaufortfrancois.github.io/samples/hdcp-detection/


xhwang-chromium:
Great! Only two minor comments:

1. The "View on GitHub" link seems to be broken.
2. On Chrome 67 it'll still be experimental. Do you want to mention how to enable the experiment?

beaufortfrancois:
> 1. The "View on GitHub" link seems to be broken.

WAI. My preview URL is a convenient way to test the API but "View on GitHub" URL is broken. It will link to the appropriate file when deployed. Thanks for flagging.

> 2. On Chrome 67 it'll still be experimental. Do you want to mention how to enable the experiment?

As you can see in https://github.com/GoogleChrome/samples/pull/574/files, we'll remove the OT part when it is promoted. For now, the present note "Note: In Chrome 67, it is available for trials so that websites can use it without any experimental flag." is enough I believe.

xhwang-chromium:
Thanks for the explanation! lgtm

raechal07 (approved):

beaufortfrancois:
@xhwang-chromium I've updated the sample as I've been able to play with it in Chrome Canary. Let me know if that still looks good to you.

xhwang-chromium:
still lgtm, thanks!

PR #575, opened 2018-04-09
Changed Presentation Receiver API sample to Chrome 59+

beaufortfrancois:
@mfoltzgoogle Can you confirm the Presentation Receiver API was added in Chrome 59 not Chrome 58?


markafoltz:
That's correct.  Here is the implementation bug:

https://bugs.chromium.org/p/chromium/issues/detail?id=525660

markafoltz (approved):

PR #582, opened 2018-05-07
[Presentation API] Update receiver sample to check connections after DOM is loaded.

markafoltz:
The connectionList promise may resolve before the DOM is loaded, which will throw when trying to update the message list in addConnection().

This moves the promise resolution into a DOMContentLoaded listener.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,7 +1,7 @@
 encoding: utf-8
 exclude: [vendor]
-gh_pages_url_prefix: https://googlechrome.github.io/samples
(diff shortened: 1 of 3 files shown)


markafoltz:
PTAL @beaufortfrancois @imcheng

markafoltz:
Merging this before I/O would be appreciated :smile:

beaufortfrancois (approved):

beaufortfrancois:
Thank you @mfoltzgoogle!

PR #534, opened 2017-08-16
Remove deprecated payment network usage. Fix #53.

dandv:

PR #585, opened 2018-05-10
Fix payment-request demos w/ regards to new specs

jonathanGB:
Current state of **PaymentRequest** demos had two problems related to the ongoing specs:

1. Use of `supportedMethods` is now a *String* rather than an *Array*
2. Re-using the same **PaymentRequest** instance when clicking on "Buy" would lead to errors (at least on FF). I have not delved too much into the causes, but my intuitition is that showing a previous **PaymentRequest** causes an error (as it wasn't part of "user activation")
image
image

Fix to **1** was trivial. 
Fix to **2** required to create a new **PaymentRequest** instance every time we wanted to show the interface. 

With the current PR, the demos now work well with the new specs.

Note: this can replace PR #534.

CC: @rsolomakhin @mnoorenberghe

### DIFF
--- a/paymentrequest/android-pay/demo.js
+++ b/paymentrequest/android-pay/demo.js
@@ -7,7 +7,7 @@
  */
 function initPaymentRequest() {
   let supportedInstruments = [{
-    supportedMethods: ['https://android.com/pay'],
+    supportedMethods: 'https://android.com/pay',
     data: {
       merchantName: 'Android Pay Demo',
       // Place your own Android Pay merchant ID here. The merchant ID is tied to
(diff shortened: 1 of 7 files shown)


jonathanGB:
I signed it!

mnoorenberghe:
> Fix to 2 required to create a new PaymentRequest instance every time we wanted to show the interface.

I suspect that is just a Fx bug and probably shouldn't be in this PR.

mnoorenberghe:
@jonathanGB Do you have time to revert the changes for part 2?

jonathanGB:
I'm gonna look at it this week!

jonathanGB:
@mnoorenberghe this should do it; only "supportedMethods" has been changed now.

mnoorenberghe (commented):
LGTM. Maybe @beaufortfrancois can review while @rsolomakhin is away?

rsolomakhin (approved):

rsolomakhin:
@agektmr Please merge.

PaulKinlan (approved):

PR #588, opened 2018-05-27
Set pip button

beaufortfrancois:
Based on latest article changes, I've updated this sample.

You can try a live preview at https://beaufortfrancois.github.io/samples/picture-in-picture/
I'd recommend setting `Slow 3G` in DevTools Network panel to observe new changes.
@mounirlamouri @japacible

### DIFF
--- a/picture-in-picture/index.js
+++ b/picture-in-picture/index.js
@@ -1,5 +1,3 @@
-togglePipButton.disabled = !document.pictureInPictureEnabled;
-
 togglePipButton.addEventListener('click', async function(event) {
   togglePipButton.disabled = true;
   try {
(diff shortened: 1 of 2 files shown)


picture-in-picture/index.js
```diff
@@ -31,3 +29,18 @@ function onPipWindowResize(event) {
...
+
+// Set button ability if Picture-in-Picture can be used.
+setPipButton();
+
+video.addEventListener('loadedmetadata', setPipButton);
+video.addEventListener('emptied', setPipButton);
+
+function setPipButton() {
+  togglePipButton.disabled = true ? (video.readyState === 0) :
+      !document.pictureInPictureEnabled || video.disablePictureInPicture;
```
mounirlamouri: `true ?` means you will only check for the `readyState` or am I missing something?

picture-in-picture/index.js
```diff
@@ -31,3 +29,18 @@ function onPipWindowResize(event) {
   const pipWindow = event.target;
   log(`> Window size is ${pipWindow.width}x${pipWindow.height}`);
 }
+
+
+// Hide button if Picture-in-Picture is not supported.
+togglePipButton.hidden = !('pictureInPictureEnabled' in document);
```
mounirlamouri: Should this be: 
```javascript
if ('pictureInPictureEnabled' in document) {
  setPipButton();
  // add listeners and all ...
} else {
  togglePipButton. hidden = true;
}
```

beaufortfrancois: Done.

picture-in-picture/index.js
beaufortfrancois: I was thinking about checking `readyState === 0` only but this ternary thing makes it harder to read. So I've removed it.

Thanks!

mounirlamouri (approved):

PR #600, opened 2018-07-24
Add sourceBuffer.changeType sample

beaufortfrancois:
Hello @wolenetz,

What do you think of this sample?
I'll mention Origin Trial when it's confirmed.

I'll also upload these files to a Google Cloud Storage instance.


wolenetz:
@beaufortfrancois No Origin Trial was started; we shipped it on-by-default in M70.
Looking at the sample now.

media/sourcebuffer-changetype.html
```diff
@@ -0,0 +1,29 @@
+---
+feature_name: Codec and container switching in MSE
+chrome_version: 70
+feature_id: 5719220952236032
+check_min_version: true
+---
+
+<h3>Background</h3>
+<p>
+Chrome is experimenting with adding support for improved cross-codec or
```
wolenetz: s/experimenting with adding/has added/

media/sourcebuffer-changetype.html
```diff
@@ -0,0 +1,29 @@
...
+</p>
+<p>
+This sample demonstrates the use of this new method by appending different
+types of media bytes to a single SourceBuffer of a MediaSource attached to an
+audio element.
+</p>
+
+<audio controls></audio>
+
+{% include output_helper.html initial_output_content=initial_output_content %}
```
wolenetz, line 23: I'm unclear what this line actually does. rubber-stamping it.

media/sourcebuffer-changetype.html
```diff
@@ -0,0 +1,29 @@
...
+
+<script>
+  log = ChromeSamples.log;
+</script>
+
+{% include js_snippet.html filename='sourcebuffer-changetype.js' %}
```
wolenetz, line 29: Ditto

media/sourcebuffer-changetype.js
```diff
@@ -0,0 +1,50 @@
+const mediaSource = new MediaSource();
+const audio = document.querySelector('audio');
+audio.src = URL.createObjectURL(mediaSource);
+
+mediaSource.addEventListener('sourceopen', function() {
+  URL.revokeObjectURL(audio.src);
+  feedAudioElement(0 /* index */);
```
wolenetz: Perhaps for readability, s/index/resources_index/ and move the "const resources = ...;" block to before this event listener?

media/sourcebuffer-changetype.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  try {
+    let sourceBuffer;
+    const resource = resources[index];
+
+    if (mediaSource.sourceBuffers.length == 0) {
+      log('Adding source buffer with type: ' + resource.type);
+      sourceBuffer = mediaSource.addSourceBuffer(resource.type);
+      sourceBuffer.mode = 'sequence';
```
wolenetz: Perhaps add comment that 'sequence' is used here purely for simplicity, and that a more sophisticated MSE app would know which timestampOffset and appendWindow to use for each appendBuffer operation when splicing media.

media/sourcebuffer-changetype.js
```diff
@@ -0,0 +1,50 @@
...
+    log('Appending data to audio element...');
+    sourceBuffer.appendBuffer(data);
+    sourceBuffer.addEventListener('updateend', function() {
+      log('Source buffer updated');
+      feedAudioElement(++index);
+    }, { once: true });
+
+  } catch(error) {
+    log('Argh! ' + error);
+    feedAudioElement(++index);
```
wolenetz: At this point, not much reliability in the outcome of feedAudioElement(++index). Error could have resulted due to lack of support for the mime type (in changeType() or addSourceBuffer().

wolenetz (changes requested):
Looks pretty good. Just a few nits.

media/sourcebuffer-changetype.html
beaufortfrancois: Done.

media/sourcebuffer-changetype.html
beaufortfrancois: This is part of the samples architecture. Please ignore.

media/sourcebuffer-changetype.html
beaufortfrancois: Ditto

media/sourcebuffer-changetype.js
beaufortfrancois: Done.

media/sourcebuffer-changetype.js
beaufortfrancois: Done.

media/sourcebuffer-changetype.js
beaufortfrancois: I think that's OK.
Goal here is to iterate through buffers and make this demo fail gracefully in Safari (with no support for webm container).

wolenetz: Ack

PR #615, opened 2018-09-21
Add pictureInPictureWindow to enterpictureinpicture event

beaufortfrancois:
@mounirlamouri Where do you think we should add resize event listener for pipWindow? 

Context: https://github.com/WICG/picture-in-picture/pull/84

### DIFF
--- a/picture-in-picture/index.html
+++ b/picture-in-picture/index.html
@@ -1,6 +1,6 @@
 ---
 feature_name: Picture-in-Picture
-chrome_version: 69
+chrome_version: 71
 feature_id: 5729206566649856
 check_min_version: true
 ---
--- a/picture-in-picture/index.js
+++ b/picture-in-picture/index.js
@@ -4,7 +4,6 @@ togglePipButton.addEventListener('click', async function(event) {
     if (video !== document.pictureInPictureElement) {
       log('Requesting Picture-in-Picture...');
       const pipWindow = await video.requestPictureInPicture();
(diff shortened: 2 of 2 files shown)


beaufortfrancois:
I'm basically wondering if we should advocate for the pattern below:

```js
async function onButtonClick() {
  myButton.disabled = false;
  await video.requestPictureInPicture();
  myButton.disabled = false;
}

let pipWindow;

video.addEventListener('enterpictureinpicture', function(event) {
  // Video entered PiP.
  pipWindow = event.pictureInPictureWindow;
  pipWindow.addEventListener('resize', onResizePictureInPictureWindow);
});

video.addEventListener('leavepictureinpicture', function() {
  // Video left PiP.
  pipWindow.removeEventListener('resize', onResizePictureInPictureWindow);
});

function onResizePictureInPictureWindow(event) {
  const pipWindow = event.target;
  // TODO: Update video dimension based on pipWindow.width and pipWindow.height
}
```

My concern is that I'm not sure if removing event listener is worth.

mounirlamouri:
I'm a bit confused. The title, the first comment and the example/code change do not quite match.

I will assume the title is due to this being a follow-up of #84. Regarding the first comment, `PictureInPictureWindow` already has a resize event so I'm guessing the question is mostly about what pattern we should recommend? I think that we should recommend the pattern that you have in the second comment. I think it's worth clarifying why this is a better pattern and mention auto-PIP being handled.

beaufortfrancois:
Thanks @mounirlamouri!

PR #619, opened 2018-11-20
Adds `controls` attribute to final controlsList media sample

bgradin:
The final sample on the controlsList samples page is missing the `controls` attribute, so you can't see the controls.


bgradin:
I signed it!

beaufortfrancois:
The last one is specifically about not showing `controls`. The previous one has `controls` though.

image

PR #624, opened 2018-12-27
Update CSS motion path demo to use current property names

ChumpChief:
Demo isn't working currently since it's using the old property names.  This change fixes that.


ChumpChief:
I signed it!

ChumpChief:
Hi @jeffposnick, would you be the right person to review/merge this change?

PR #629, opened 2019-01-31
Update HDCP sample

beaufortfrancois:
This PR is about promoting the HDCP Policy Check API to stable and removing old enums.

TBR @xhwang-chromium


beaufortfrancois:
Live at https://googlechrome.github.io/samples/hdcp-detection/

PR #640, opened 2019-05-03
Add CSS pseudo-class :picture-in-picture Sample

beaufortfrancois:
Here's the sample for CSS pseudo-class `:picture-in-picture`.

image

R: @mounirlamouri

### DIFF
--- a/picture-in-picture/css-pseudo-class.html
+++ b/picture-in-picture/css-pseudo-class.html
@@ -1,0 +1,70 @@
+---
+feature_name: CSS pseudo-class ":picture-in-picture"
+chrome_version: 76
+feature_id: 6259534815297536
+check_min_version: true
+---
+
+<h3>Background</h3>
+<p>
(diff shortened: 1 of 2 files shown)


picture-in-picture/css-pseudo-class.html
```diff
@@ -0,0 +1,70 @@
...
+</style>
+
+{% capture css %}
+#video {
+  display: block;
+  width: 100%;
+  will-change: opacity;
+}
+
+#video:not([controls]):picture-in-picture {
```
mounirlamouri, line 32: is :not([controls]) needed?

mounirlamouri (approved):

picture-in-picture/css-pseudo-class.html
beaufortfrancois: It is not needed per se, but it is good practice to not hide video with native controls in Picture-in-Picture as this is the only way for now to navigate in video.

PR #645, opened 2019-08-05
Add stop media session action

beaufortfrancois:
FYI @mounirlamouri @beccahughes


beaufortfrancois:
https://googlechrome.github.io/samples/media-session/video.html and https://googlechrome.github.io/samples/media-session/audio.html samples are now updated with "stop" media session action.

PR #650, opened 2019-09-25
Update seeking in Media Session samples

beaufortfrancois:
This PR is about setting a seekto media session action handler and using new seekOffset in seek media session action handler.

@mounirlamouri Does that look good?

Live preview: 
https://beaufortfrancois.github.io/samples/media-session/audio.html
https://beaufortfrancois.github.io/samples/media-session/video.html

### DIFF
--- a/media-session/audio.js
+++ b/media-session/audio.js
@@ -48,15 +48,17 @@ audio.addEventListener('ended', function() {
 
 /* Seek Backward & Seek Forward */
 
-let skipTime = 10; /* Time to skip in seconds */
+let defaultSkipTime = 10; /* Time to skip in seconds by default */
 
-navigator.mediaSession.setActionHandler('seekbackward', function() {
+navigator.mediaSession.setActionHandler('seekbackward', function(event) {
   log('> User clicked "Seek Backward" icon.');
+  const skipTime = event.seekOffset || defaultSkipTime;
   audio.currentTime = Math.max(audio.currentTime - skipTime, 0);
 });
 
-navigator.mediaSession.setActionHandler('seekforward', function() {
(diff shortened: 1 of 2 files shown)


media-session/audio.js
```diff
@@ -85,6 +87,20 @@ try {
...
 }
 
+/* Seek To (supported since Chrome 78) */
+
+try {
+  navigator.mediaSession.setActionHandler('seekto', function(event) {
+    log('> User clicked "Seek To" icon.');
+    if (event.fastSeek) {
+      return;
+    }
```
mounirlamouri, line 97: Why bailing if `fastSeek` is set?

beaufortfrancois: I thought it would be called again. What would you do instead?

mounirlamouri: I don't really have a recommendation. I think `fastSeek` is when you scrub the timeline instead of when you tab on it. I am not sure it will end with a non-fastSeek call at the end. I don't think Chrome will ever send something with `fastSeek = true` so I can't give more insight here.

Maybe this could be linked to the fast seek call in the HTMLMediaElement?

beaufortfrancois: Shall I simply omit then from this sample until we figure out what to do with it?

mounirlamouri: I think we should do that:
```
if (event.fastSeek && ('fastSeek' in audio)) {
  audio.fastSeek(event.seekTime);
}
```

This way, we ignore it on Chrome but not on browsers that implement `fastSeek()`. I want to be mindful of them.

beaufortfrancois: Thanks!

PR #648, opened 2019-09-04
Add Web Bluetooth Availability Sample

beaufortfrancois:
This PR is adding two samples for [Web Bluetooth getAvailability()](https://www.chromestatus.com/features/6229592081694720)

Does that look good @odejesush?

Live preview:
https://beaufortfrancois.github.io/samples/web-bluetooth/availability.html
https://beaufortfrancois.github.io/samples/web-bluetooth/availability-async-await.html
https://beaufortfrancois.github.io/samples/web-bluetooth/index.html

FYI @jyasskin @g-ortuno @scheib

PR #651, opened 2019-09-30
Web Bluetooth Scanning Sample, initial draft.

scheib:
Introducing the most basic of scanning samples.  Intent is to show
that scanning can be started, demonstrate the permissions UI,
and that advertisement packets are received.

Several details are omitted.  E.g. filtering by service data,
and showing advertisement packet contents.

### DIFF
--- a/web-bluetooth/scan.html
+++ b/web-bluetooth/scan.html
@@ -1,0 +1,39 @@
+---
+feature_name: Web Bluetooth / Scanning (Async Await)
+chrome_version: 79
+check_min_version: true
+feature_id: 5346724402954240
+icon_url: icon.png
+index: index.html
+---
+
+{% include_relative _includes/intro.html %}
+
+<p>This sample illustrates basic use of the Web Bluetooth Scanning API to
+report advertising packets from nearby Bluetooth Low Energy Devices.</p>
+
+<form>
+  <label for="allAdvertisements">All Advertisements</label>
+  <input id="allAdvertisements" type="checkbox">
+  <input id="name" type="text" size=17 placeholder="Device Name">
+  <input id="namePrefix" type="text" size=17 placeholder="Device Name Prefix">
+  <button>Scan for Bluetooth Advertisements</button>
+</form>
+
+{% include output_helper.html %}
+
+{% include js_snippet.html filename='scan.js' %}
+
+<script>
+  document.querySelector('form').addEventListener('submit', function(event) {
+    event.stopPropagation();
+    event.preventDefault();
+
+    if (isWebBluetoothEnabled()) {
+      ChromeSamples.clearLog();
+      onButtonClick();
+    }
+  });
+</script>
+
+{% include_relative _includes/utils.html %}
--- a/web-bluetooth/scan.html
+++ b/web-bluetooth/scan.html
@@ -12,6 +12,10 @@ index: index.html
 <p>This sample illustrates basic use of the Web Bluetooth Scanning API to
 report advertising packets from nearby Bluetooth Low Energy Devices.</p>
 
+<p>Note: Scanning is still under development. You must be using Chrome 79+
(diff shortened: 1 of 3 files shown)


scheib:
@dougt, @beaufortfrancois, Please take a look for review.

web-bluetooth/scan.js
```diff
@@ -0,0 +1,34 @@
...
+  if (document.querySelector('#allAdvertisements').checked) {
+    options.acceptAllAdvertisements = true;
+  } else {
+    options.filters = filters;
+  }
+
+  try {
+    log('Requesting Bluetooth Scan...');
+    log('with ' + JSON.stringify(options));
+    await navigator.bluetooth.requestLEScan(options);
```
beaufortfrancois: I think basic demo should at least allow users to stop scanning.
How about something like below?
```js
var scan;

async function onStartButtonClick() {
  /* ... */
  scan = await navigator.bluetooth.requestLEScan(options);
  /* ... */
}

async function onStopButtonClick() {
  if (!scan || !scan.active) {
    return;
  }
  scan.stop();
}
`

File: web-bluetooth/scan.js [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+
+    log('Scan started...');
```
beaufortfrancois: Shall we show scan options?

beaufortfrancois (commented):
Thank you for writing this. I left some comments
I'm happy to write following ones ;)

beaufortfrancois:
@scheib See https://github.com/GoogleChrome/samples/pull/648

By the way, we may want to add a "promise" test as well.

web-bluetooth/scan.js
scheib: Done

web-bluetooth/scan.js
scheib: I'd like to keep the code as simple as possible, I've used a simple timer to stop the scan after 10 seconds.

web-bluetooth/scan.js
scheib: done

scheib:
> @scheib See #648
> 
> By the way, we may want to add a "promise" test as well.

I suggest we no longer create pairs, since it's been so many years and async-await is now pervasive https://caniuse.com/#feat=async-functions.  Especially if we're creating a new batch of scanning related samples, I don't see the need.

web-bluetooth/scan.html
```diff
@@ -0,0 +1,39 @@
+---
+feature_name: Web Bluetooth / Scanning
+chrome_version: 79
```
beaufortfrancois, line 3: Let's make sure to update this if feature gets postponed.

web-bluetooth/scan.js
```diff
@@ -0,0 +1,44 @@
...
+    const scan = await navigator.bluetooth.requestLEScan(options);
+
+    log('Scan started with:');
+    log(' acceptAllAdvertisements: ' + scan.acceptAllAdvertisements);
+    log(' active: ' + scan.active);
+    log(' keepRepeatedDevices: ' + scan.keepRepeatedDevices);
+    log(' filters: ' + JSON.stringify(scan.filters));
+
+    navigator.bluetooth.addEventListener('advertisementreceived', event => {
+      log("Advertisement received.");
```
beaufortfrancois: Nit: Use simple quotes for consistency.

beaufortfrancois (approved):

web-bluetooth/scan.html
scheib: OK, also added note that chrome://flags/#enable-experimental-web-platform-features is required now.

beaufortfrancois:
Sample is now live at https://googlechrome.github.io/samples/web-bluetooth/scan.html

beaufortfrancois:
I've added it to https://www.chromestatus.com/feature/5346724402954240

PR #654, opened 2019-10-08
Web Share (Target) Level 2 sample PWA

jeffposnick:
CC: @Snugug

An offline-capable PWA that illustrates using the Web Share (Target) Level 2 APIs.

This is the Svelte source code + build process. It's currently deployed at https://stoic-kepler-c046f7.netlify.com, but we will likely find a more official deployment URL.


jeffposnick:
@googlebot I fixed it.

web-share/README.md
```diff
@@ -0,0 +1,57 @@
...
+All media files are stored locally in the browser using the
+[Cache Storage API](https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/cache-api).
+Media shared to this web app is _not_ backed up anywhere—please keep this in
+mind before deleting something. Anything you'd like to keep should be shared to
+a more permanent location.
+
+## Requirements
+
+To fully use this web app, you need a combination of browser and operating
+system that supports the underlying APIs. As of Sept. 2019, that's limited to
```
beaufortfrancois: Nit: October

web-share/README.md
```diff
@@ -0,0 +1,57 @@
...
+Chrome on Android.
+
+Before you can receive incoming media files via the Web Share Target API Level
+2, you need to add the web app to your home screen. This will register it as a
+share target with the operating system, and you'll start seeing it in the list
+of available share targets when you attempt to share supported media from a
```
beaufortfrancois, line 27: Shall we include screenshots?

web-share/src/svelte/routes/Audio.svelte
```diff
@@ -0,0 +1,40 @@
...
+	export const svg = '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M0 0h24v24H0z" fill="none"/><path d="M12 3v9.28c-.47-.17-.97-.28-1.5-.28C8.01 12 6 14.01 6 16.5S8.01 21 10.5 21c2.31 0 4.2-1.75 4.45-4H15V6h4V3h-7z"/></svg>';
+  export const text = 'Audio';
+	export const title = 'Listen to saved audio.';
+</script>
+
+<script>
+  import {fade} from 'svelte/transition';
+
+  import {fadeConfig} from '../../js/constants';
+	import {getCachedMediaMetadata} from '../../js/getCachedMediaMetadata';
```
beaufortfrancois: Nit: align import

web-share/src/svelte/routes/Images.svelte
```diff
@@ -0,0 +1,40 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<script context="module">
+  export const href = '/images';
+  export const mimePrefix = 'image/';
+  export const mimeRoute = {href, mimePrefix};
+	export const svg = '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M0 0h24v24H0z" fill="none"/><path d="M21 19V5c0-1.1-.9-2-2-2H5c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h14c1.1 0 2-.9 2-2zM8.5 13.5l2.5 3.01L14.5 12l4.5 6H5l3.5-4.5z"/></svg>';
```
beaufortfrancois: Nit: align export(s)

web-share/src/js/service-worker.js
```diff
@@ -0,0 +1,82 @@
...
+  for (const mediaFile of mediaFiles) {
+    // TODO: Instead of bailing, come up with a
+    // default name for each possible MIME type.
+    if (!mediaFile.name) {
+      if (broadcastChannel) {
+        broadcastChannel.postMessage('Sorry! No name found on incoming media.');
+      }
+      continue;
+    }
+    await cache.put(
```
beaufortfrancois, line 32: Why await here?
You could add every `cache.put` promise and wait with `Promise.all` at the end right?

web-share/src/js/getCachedMediaMetadata.js
```diff
@@ -0,0 +1,36 @@
...
+    return {
+      contentType: response.headers.get('content-type'),
+      src: request.url,
+    };
+  }));
+}
+const cachedMediaMetadataPromise = _getCachedMediaMetadata();
+
+export async function getCachedMediaMetadata(contentTypePrefix) {
+  const cachedMetadata = await cachedMediaMetadataPromise;
```
beaufortfrancois, line 33: I think* it means `cachedMetadata` will **always** return the same cached media metadata
```suggestion
  const cachedMetadata = await _getCachedMediaMetadata();
```

web-share/src/svelte/components/Media.svelte
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	async function handleDelete() {
+		const cache = await caches.open('media');
+		await cache.delete(src);
+		window.location.reload();
+	}
+
+	async function handleShare() {
+		const cache = await caches.open('media');
```
beaufortfrancois: Should this be `cacheName` as well here?

web-share/src/svelte/components/MediaList.svelte
```diff
@@ -0,0 +1,48 @@
...
+{#if cachedMediaMetadatas.length > 0}
+	<div>
+		{#each cachedMediaMetadatas as metadata}
+			<Media {...metadata}/>
+		{/each}
+	</div>
+{:else}
+	<p>You don't have any saved {typeOfMedia}.</p>
+	<ol>
+		<li><a href="https://developers.google.com/web/fundamentals/app-install-banners/">Add</a> this web app to your homescreen on Android, using Chrome 76+.</li>
```
beaufortfrancois: Shall we skip this step if user has already installed web app?

web-share/src/svelte/components/Media.svelte
```diff
@@ -0,0 +1,94 @@
...
+	audio, img, video {
+		max-width: 100%;
+	}
+</style>
+
+<div class="card">
+	{#if contentType.startsWith('image/')}
+		<img {src} alt="An image in the scrapbook.">
+	{:else if contentType.startsWith('video/')}
+		<video {src} controls alt="A video in the scrapbook."></video>
```
beaufortfrancois: Issue: When downloading audio and video from media controls, I get "No file" in media notification. Is that intended?

beaufortfrancois: If so, you may want to set `controlsList` to `nodownload`: https://developers.google.com/web/updates/2017/03/chrome-58-media-updates#controlslist

web-share/src/svelte/routes/Videos.svelte
```diff
@@ -0,0 +1,41 @@
+<!--
```
beaufortfrancois, line 1: For consistency, you may want to rename this file to `Video.svelte`.
Same would apply to `Image.svelte`

beaufortfrancois (commented):
@jeffposnick Great app!

web-share/src/svelte/components/Media.svelte
beaufortfrancois: I noticed it only happens with Google Photos. When "sending file" with the Files app. It works seamlessly.

web-share/src/svelte/components/Media.svelte
```diff
@@ -0,0 +1,94 @@
...
+<script>
+	import Button from './Button.svelte';
+
+	export let contentType = '';
+	export let src = '';
+
+	const canShare = 'canShare' in navigator;
+
+	function handleView() {
+		window.location.href = src;
```
beaufortfrancois: For audio/video files, it downloads file while image shows it inline. Is that intended?

web-share/src/svelte/routes/Audio.svelte
jeffposnick: Yeah, I developed some of this project using an IDE that apparently was set for tabs instead of spaces...

web-share/src/svelte/routes/Videos.svelte
jeffposnick: My thinking was to have the filename match the `href`, which use `/videos` (and `/images`). `/audio` is kind of the outlier when it comes to pluralization.

web-share/src/svelte/components/Media.svelte
jeffposnick: No, it wasn't really. I was just looking for a quick way of showing the image in a standalone view, and didn't realize it wouldn't also work for audio and video.

I can add a separate `View.svelte` component that will display any cached media file inline.

web-share/src/svelte/components/Media.svelte
jeffposnick: Thanks! I didn't know about `controlsList="nodownload"`. I think adding that makes sense vs. figuring out why some media files will download and others won't.

web-share/src/js/service-worker.js
jeffposnick: I didn't _put_ a huge amount of thought into it, but I guess the current approach has two advantages, given that each `mediaFile` might potentially be quite large:

- Caching one-by-one seems like it would play nicer if our PWA is close to using up their storage quota on a given device.

- I could imagine that streaming the `mediaFile` bodies into new `Response` objects in the course of writing them to cache storage could be memory intensive, and doing that one at a time instead of a bunch in parallel seems less likely to degrade performance.

But... like I said, it's not something that I really thought about ahead of time.

(Before `async`/`await`, when everything had to be chained together via `.then()`, I was more inclined to use things like `.map()` + `Promise.all()`, but I do that less now as a matter of style.)

web-share/src/js/getCachedMediaMetadata.js
jeffposnick: Yeah, because this is *not* a single page app, and any updates to cache storage trigger a reload, I assume that whatever the set of cached media files are at the time the page loads isn't going to change.

web-share/src/svelte/components/Media.svelte
jeffposnick: Yup!

web-share/src/svelte/components/MediaList.svelte
jeffposnick: Yeah, I put in some conditional logic based on https://developers.google.com/web/fundamentals/app-install-banners#detect-mode

web-share/src/svelte/components/Media.svelte
beaufortfrancois: If not too hard, that would be great.

web-share/src/svelte/routes/Videos.svelte
beaufortfrancois: Noted.

web-share/src/svelte/components/MediaList.svelte
beaufortfrancois: Cool!

web-share/src/js/getCachedMediaMetadata.js
beaufortfrancois: Oki

web-share/src/svelte/components/Media.svelte
jeffposnick: Yup! Should be able to get to this over the next few days.

web-share/src/js/service-worker.js
beaufortfrancois: Noted.

web-share/README.md
beaufortfrancois: Or not?

beaufortfrancois (approved):
LGTM with nits.

web-share/README.md
jeffposnick: They might change over time, but sure, I can include some to start.

web-share/src/svelte/components/Media.svelte
jeffposnick: Added. I might need to tweak the UI a bit, but `/view/...` routes should work now.

web-share/README.md
jeffposnick: Actually, let me hold off for now, as I do think we're going to get some UI changes in.

PR #657, opened 2019-10-25
Custom offline page service worker revamp

jeffposnick:
I still refer folks to this periodically, and I realize that it could really use a revamp; modern JavaScript, and, in particular, illustrate the use of navigation preload when the browser supports it.


beaufortfrancois (approved):
LGTM

service-worker/custom-offline-page/service-worker.js
```diff
@@ -1,5 +1,5 @@
 /*
- Copyright 2015 Google Inc. All Rights Reserved.
+ Copyright 2019 Google Inc. All Rights Reserved.
```
tomayac: If you update the copyright year, you give up the copyright before the updated year. If the code has substantially changed, the recommendation is to add a new year:

 ```suggestion
 Copyright 2015, 2019 Google Inc. All Rights Reserved.
```

File: service-worker/custom-offline-page/service-worker.js [hunk 2]
```diff
@@ -11,92 +11,66 @@
...
-    fetch(createCacheBustedRequest(OFFLINE_URL)).then(function(response) {
-      return caches.open(CURRENT_CACHES.offline).then(function(cache) {
-        return cache.put(OFFLINE_URL, response);
-      });
-    })
-  );
+self.addEventListener('install', (event) => {
+  event.waitUntil((async () => {
+    const cache = await caches.open(CACHE_NAME);
+    // Setting cache: reload in the new request will ensure that the response
```
tomayac: The 'reload' value is a string: https://developer.mozilla.org/en-US/docs/Web/API/Request/cache#Value.

```suggestion
    // Setting {cache: 'reload'} in the new request will ensure that the response
```

service-worker/custom-offline-page/service-worker.js
```diff
@@ -11,92 +11,66 @@
...
+        if (preloadResponse) {
+          return preloadResponse;
+        }
+
+        const networkResponse = await fetch(event.request);
+        return networkResponse;
+      } catch (error) {
+        // catch is only triggered if an exception is thrown, which is likely
+        // due to a network error.
+        // If fetch() returns a valid HTTP response with an response code in
```
tomayac: Typo:

```suggestion
        // If fetch() returns a valid HTTP response with a response code in
```

tomayac (approved):
LGTM2, with three minor nits.

service-worker/custom-offline-page/service-worker.js
jeffposnick: Gotcha.

PR #659, opened 2019-11-05
Add more detail to Bluetooth scanning example

tomayac:
Reads out some of the advertisement data and logs it: 

- Current logging:

  ```
  Advertisement received.
  ```

- New logging: 
  ```
  Advertisement received.
    Device ID: TcgbPzvpvDFznj7QE8n02w==
    RSSI: -66
    TX Power: null
    UUIDs: 
    Manufacturer Data: 135
    Service Data: 00003e10-0000-1000-8000-00805f9b34fb
  ```

CC: @scheib

### DIFF
--- a/web-bluetooth/scan.js
+++ b/web-bluetooth/scan.js
@@ -30,15 +30,31 @@ async function onButtonClick() {
 
     navigator.bluetooth.addEventListener('advertisementreceived', event => {
       log('Advertisement received.');
+      log('Device Name: ' + event.device.name);
       log('  Device ID: ' + event.device.id);
       log('  RSSI: ' + event.rssi);
       log('  TX Power: ' + event.txPower);
       log('  UUIDs: ' + event.uuids);
-      event.manufacturerData.forEach((_, key) => {
-        log('  Manufacturer Data: ' + key);
+
+      const valueDataLogger = (key, valueDataView, whatData) => {
+        const uInt8Array = [...new Uint8Array(valueDataView.buffer)];
+        const hexString = uInt8Array.map(b => {
+          return b.toString(16).padStart(2, '0');
+        }).join('');
+        const asciiString = hexString.match(/.{1,2}/g).reduce((acc,char) => {
+          return acc + String.fromCharCode(parseInt(char, 16))
+        }, '');
+        log(`  ${whatData} Data: ` + key +
+            '\n    (Raw) ' + uInt8Array +
+            '\n    (Hex) ' + hexString +
+            '\n    (ASCII) ' + asciiString);
+      };
+
+      event.manufacturerData.forEach((valueDataView, key) => {
+        valueDataLogger(key, valueDataView, 'Manufacturer');
       });
-      event.serviceData.forEach((_, key) => {
-        log('  Service Data: ' + key);
+      event.serviceData.forEach((valueDataView, key) => {
+        valueDataLogger(key, valueDataView, 'Service');
       });
     });
 
--- a/web-bluetooth/scan.js
+++ b/web-bluetooth/scan.js
@@ -30,6 +30,16 @@ async function onButtonClick() {
 
     navigator.bluetooth.addEventListener('advertisementreceived', event => {
       log('Advertisement received.');
+      log('  Device ID: ' + event.device.id);
+      log('  RSSI: ' + event.rssi);
+      log('  TX Power: ' + event.txPower);
+      log('  UUIDs: ' + event.uuids);
+      event.manufacturerData.forEach((_, key) => {
+        log('  Manufacturer Data: ' + key);
+      });
+      event.serviceData.forEach((_, key) => {
+        log('  Service Data: ' + key);
+      });
     });
 
     setTimeout(stopScan, 10000);
--- a/web-bluetooth/scan.js
+++ b/web-bluetooth/scan.js
@@ -30,17 +30,16 @@ async function onButtonClick() {
 
     navigator.bluetooth.addEventListener('advertisementreceived', event => {
       log('Advertisement received.');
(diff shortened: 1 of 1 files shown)


web-bluetooth/scan.js
```diff
@@ -30,6 +30,16 @@ async function onButtonClick() {
...
+      log('  Device ID: ' + event.device.id);
+      log('  RSSI: ' + event.rssi);
+      log('  TX Power: ' + event.txPower);
+      log('  UUIDs: ' + event.uuids);
+      event.manufacturerData.forEach((_, key) => {
+        log('  Manufacturer Data: ' + key);
+      });
+      event.serviceData.forEach((_, key) => {
+        log('  Service Data: ' + key);
+      });
```
scheib: If dumping the keys, we might as well read out the value bytes as well.  (They'll never get too long, as ad packets even in Bluetooth 5 is 255 bytes, but most devices will just be less than a dozen.)

Something like:

...Data.forEach((valueDataView, key) => {
log.' ...Data: ' + key + ': ' + (new Uint8Array(valueDataView.buffer).toString()

That should get things out in decimal.. a tad nicer might be to print in hex https://stackoverflow.com/questions/40031688/javascript-arraybuffer-to-hex

Super fancy would be an option drop down in the sample that chose to print as decimal, hex, or ASCII characters.

tomayac:
Thanks, @scheib. I have now added Raw, Hex, and ASCII logging, which seems to work:

```
Advertisement received.
Device Name: null
  Device ID: XxYaCuux5dTzXsEm+XMbGg==
  RSSI: -54
  TX Power: null
  UUIDs: 0000feaa-0000-1000-8000-00805f9b34fb
  Service Data: 0000feaa-0000-1000-8000-00805f9b34fb
    (Raw) 16,238,3,103,111,111,46,103,108,47,112,73,87,100,105,114
    (Hex) 10ee03676f6f2e676c2f704957646972
    (ASCII) îgoo.gl/pIWdir
```

This is from my Thingy:52 device, and the ASCII value `îgoo.gl/pIWdir` contains a human-readable URL [goo.gl/pIWdir](https://goo.gl/pIWdir) that the device indeed emits. I'm not entirely sure about the gibberish before the URL though, and if this approach would work in the general case. Full disclosure: I'm outside of my comfort zone here…

tomayac:
One way to easily filter this would be to wrap it in spans:

```html
    <span class="raw">(Raw) 16,238,3,103,111,111,46,103,108,47,112,73,87,100,105,114</span>
    <span class="hex">(Hex) 10ee03676f6f2e676c2f704957646972</span>
    <span class="ascii">(ASCII) �î�goo.gl/pIWdir</span>
```

The output filter selector could then just trigger the `hidden` attribute of each class.

web-bluetooth/scan.js
```diff
@@ -30,6 +30,32 @@ async function onButtonClick() {
 
     navigator.bluetooth.addEventListener('advertisementreceived', event => {
       log('Advertisement received.');
+      log('Device Name: ' + event.device.name);
+      log('  Device ID: ' + event.device.id);
+      log('  RSSI: ' + event.rssi);
+      log('  TX Power: ' + event.txPower);
+      log('  UUIDs: ' + event.uuids);
+
+      const valueDataLogger = (key, valueDataView, whatData) => {
```
beaufortfrancois: Please move this at the very bottom in `/* utils */`.
See https://googlechrome.github.io/samples/web-bluetooth/device-information-characteristics.html for instance.

web-bluetooth/scan.js
```diff
@@ -30,6 +30,32 @@ async function onButtonClick() {
...
+        const uInt8Array = [...new Uint8Array(valueDataView.buffer)];
+        const hexString = uInt8Array.map(b => {
```
beaufortfrancois: I'd use only hexString to be honest.
Something like below should be enough:
```js
        let a = [];
        for (let i = 0; i < valueDataView.buffer.byteLength; i++) {
          a.push(
            "0x" + ("00" + valueDataView.getUint8(i).toString(16)).slice(-2)
          );
        }
        log(`  > Data: ${key}: ${a.join(" ")}`);
```

web-bluetooth/scan.js
tomayac: Done

web-bluetooth/scan.js
```diff
@@ -42,3 +54,16 @@ async function onButtonClick() {
     log('Argh! ' + error);
   }
 }
+
+/* Utils */
+
+const valueDataLogger = (key, valueDataView, whatData) => {
+  const hexString = [...new Uint8Array(valueDataView.buffer)].map(b => {
+    return b.toString(16).padStart(2, '0');
+  }).join('');
```
scheib: I recommend `join(' ')` so that the bytes can be visually parsed.

web-bluetooth/scan.js
```diff
@@ -42,3 +54,16 @@ async function onButtonClick() {
     log('Argh! ' + error);
   }
 }
+
+/* Utils */
+
+const valueDataLogger = (key, valueDataView, whatData) => {
```
scheib: 'valueDataLogger' and 'whatData' where tricky for me to understand till I read the implementation.  May I suggest:

logDataView(labelOfDataSource, key, valueDataView)

called as logDataView('Service', key, valueDataView)

scheib (approved):
LGTM minor nits

web-bluetooth/scan.js
tomayac: Done.

web-bluetooth/scan.js
tomayac: Done.

PR #666, opened 2019-12-11
Content Indexing API updates for Web Share sample

jeffposnick:
R: @beaufortfrancois @tomayac 
CC: @snugug @rayankans

This updates the existing Web Share (Target) Level 2 demo to include an Origin Trial token and usage of the upcoming Content Indexing API.

I don't think a standalone demo for Content Indexing makes sense, as it works better as part of a larger web app. I'll link to this repo from some materials we're putting out for Chrome 80.

There's a new, "official" deployment URL that's associated with the Origin Trial: https://scrapbook-pwa.web.app/


jeffposnick:
@googlebot I fixed it.

web-share/src/js/contentIndexing.js
```diff
@@ -0,0 +1,76 @@
...
+  const allEntries = await registration.index.getAll();
+  const idsInIndex = new Set(allEntries.map((entry) => entry.id));
+
+  // Get a list of all cached media.
+  const cachedMediaMetadata = await cachedMediaMetadataPromise;
+
+  for (const metadata of cachedMediaMetadata) {
+    if (idsInIndex.has(metadata.src)) {
+      // If a given id/URL is in both the content index and currently in our
+      // cached, remove it from the set.
```
tomayac:
```suggestion
      // cache, remove it from the set.
```

web-share/src/js/contentIndexing.js
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      // Otherwise, if the id/URL is cached but not currently in the index, add
+      // it to the index.
+      // category should end up being 'image', 'audio', or 'video'.
+      const [category] = metadata.contentType.split('/');
+      await registration.index.add({
+        // Use the cached media URL as the id.
+        id: metadata.src,
+        // Our web app has a route for viewing a specific cached media URL.
+        // Note that this needs to be the URL for an HTML that will display the
```
tomayac:
```suggestion
        // Note that this needs to be the URL for a page that will display the
```

tomayac (approved):
LGTM. Approve with absolutely minor nits about comments.

Also: a typo ("an another") in [MediaList.svelte](https://github.com/jeffposnick/samples/blob/70bdc7ed69b8dc31cc0baf22812a591e55ee483e/web-share/src/svelte/components/MediaList.svelte#L49) that wasn't part of the PR.

web-share/src/js/app.js
```diff
@@ -15,6 +15,7 @@
  **/
 
 import App from '../svelte/App.svelte';
+import {syncContentIndex} from './contentIndexing.js';
```
beaufortfrancois, line 18: Shall we load this dynamically only when service Worker are supported?

web-share/src/js/contentIndexing.js
```diff
@@ -0,0 +1,76 @@
...
+  const idsInIndex = new Set(allEntries.map((entry) => entry.id));
+
+  // Get a list of all cached media.
+  const cachedMediaMetadata = await cachedMediaMetadataPromise;
+
+  for (const metadata of cachedMediaMetadata) {
+    if (idsInIndex.has(metadata.src)) {
+      // If a given id/URL is in both the content index and currently in our
+      // cached, remove it from the set.
+      idsInIndex.delete(metadata.src);
```
beaufortfrancois, line 40: maybe use `continue` to avoid indentation and make code easier to read.

web-share/src/index.html
```diff
@@ -10,6 +10,9 @@
     <link rel="stylesheet" href="css/svelte.css">
     <script defer src="js/app.js"></script>
     <link rel="manifest" href="manifest.json">
+    <!-- This Origin Trial token for Content Indexing will only work when deployed to https://scrapbook-pwa.web.app/ -->
```
beaufortfrancois, line 13: I've tried https://scrapbook-pwa.web.app and `registration.index` is `undefined`. Is this intended?

web-share/src/svelte/components/Media.svelte
```diff
@@ -96,4 +106,7 @@ limitations under the License.
       <Button handleClick={handleDelete}>Delete</Button>
     </div>
   {/if}
+  {#if showSize}
+    <div class="size">{prettyBytes(size)}</div>
```
beaufortfrancois, line 110: FYI, even though your web app says 1.98 MB, Chrome Downloads says 0B.

web-share/src/js/service-worker.js
```diff
@@ -11,6 +11,15 @@ import {mimeRoute as videosRoute} from '../svelte/routes/Videos.svelte';
 
 const broadcastChannel = 'BroadcastChannel' in self ? new BroadcastChannel(channelName) : null;
 
+self.addEventListener('contentdelete', async (event) => {
```
beaufortfrancois: Why async?

beaufortfrancois: Maybe a comment explaining that this event is fired when a user (or browser) has deleted some content?

web-share/README.md
```diff
@@ -1,14 +1,19 @@
...
-demo that shows off the use of two emerging web platform features:
+demo that shows off the use of three emerging web platform features:
 
 - [Web Share Target API Level 2](https://developers.google.com/web/updates/2018/12/web-share-target),
 for receiving incoming media files shared **by** other installed applications.
 
 - [Web Share Level 2](https://developers.google.com/web/updates/2019/05/web-share-files),
 for taking local media files and natively sharing them **to** other installed applications.
 
+- [Content Indexing API](https://chromestatus.com/feature/5658416729030656),
```
beaufortfrancois, line 14: A WIP screenshot of Chrome Downloads would really help to understand quickly what this is about.

beaufortfrancois (commented):
I've played a bit with the app (in Canary with experimental-web-platform-features flag enabled) and I've failed to share a video file, an audio file, and several image files at once. Is that something you've experienced as well?
Note that it works great with a single image file ;)

web-share/src/js/app.js
tomayac:
```js
if ('index' in await navigator.serviceWorker.ready) {}
```

Seems to be the correct way to feature-detect this.

web-share/src/index.html
tomayac: Could it simply be that this latest version is not deployed yet?

tomayac:
> I've failed to share a video file, an audio file, and several image files at once.

I didn't try multiple files to be honest, but sharing single files worked.

web-share/src/js/app.js
beaufortfrancois: Yes indeed.

web-share/src/js/service-worker.js
jeffposnick: Right, it's not needed.

web-share/README.md
jeffposnick: I'm going to switch this link out to the web.dev article (which will have screenshots) once it's live. That will be in a separate PR.

web-share/src/svelte/components/Media.svelte
jeffposnick: Yeah, the folks responsible for Chrome Downloads are aware that it's always going to be `0B` there. I think they might remove it from the Chrome UI?

Not much to do about that here, though.

web-share/src/js/service-worker.js
jeffposnick: Done.

web-share/src/index.html
jeffposnick: You need Chrome 80+ on Android.

I just confirmed that with the current Canary on Android (M81) and the origin trial token is working for https://scrapbook-pwa.web.app/

web-share/src/js/contentIndexing.js
jeffposnick: I'd prefer not to? Both clauses are equally possible and important, and I'd rather just have an if/else.

web-share/src/js/app.js
jeffposnick: It's going to be a pain to get things set up that way, given that I need to use this file to bootstrap Svelte, and I am not really sure how to do that with anything other than `iife` output from Rollup. Maybe I will figure it out down the line!

jeffposnick:
I've found that sharing movies and audio can take up to a minute or so to process, even on a fast network.

If the web app is open an running, then it should display "Saving media locally..." message when this is happening:

https://github.com/GoogleChrome/samples/blob/be99cc41fbf4f6ce14646adaa0c3d292a28f5f2f/web-share/src/js/service-worker.js#L27-L29

Unfortunately, if the incoming `POST` wakes up the web app, you'll be stuck on the splash screen until the `fetch` handler completes and it triggers a redirect, meaning that you can't display a "Please wait" message.

beaufortfrancois:
Oh! So that's why! Thanks!

PR #670, opened 2020-01-30
Add basic Web NFC sample

beaufortfrancois:
I'll push this basic Web NFC sample now and will update it if needed afterwards so that people can give it a try easily.

### DIFF
--- a/web-nfc/index.html
+++ b/web-nfc/index.html
@@ -1,0 +1,27 @@
+---
+feature_name: Web NFC
+chrome_version: 81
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
@zolkis @kenchris @riju Let me know if https://googlechrome.github.io/samples/web-nfc/ looks good to you.

beaufortfrancois:
I think we should handle basic text and URL records when scanning. Not necessarily the entire set of records type.

PR #672, opened 2020-02-06
Add position state support

beaufortfrancois:
This PR adds support for Media Session position state, coming in Chrome 81 beta.
@beccahughes @mounirlamouri please review

Live preview at https://beaufortfrancois.github.io/samples/media-session/video.html and https://beaufortfrancois.github.io/samples/media-session/audio.html

### DIFF
--- a/media-session/audio.js
+++ b/media-session/audio.js
@@ -24,6 +24,26 @@ function updateMetadata() {
     album: track.album,
     artwork: track.artwork
   });
+
+  // Media is loaded, set the duration.
+  updatePositionState();
+}
+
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
For info, I've stumbled upon a bug at https://bugs.chromium.org/p/chromium/issues/detail?id=1049530

beaufortfrancois:
If `setPositionState` is not meant to be called with `null` when media is paused, then we have a bug in Chrome though. See https://bugs.chromium.org/p/chromium/issues/detail?id=1049530

beccahughes:
LGTM

PR #678, opened 2020-03-12
Check all entries in callback

szager-chromium:
The existing sample code can cause IntersectionObserver events to be missed, because it only checks the first entry in the list. This came to light in a chromium bug:

https://bugs.chromium.org/p/chromium/issues/detail?id=941681#c8

### DIFF
--- a/intersectionobserver/demo.js
+++ b/intersectionobserver/demo.js
@@ -13,16 +13,24 @@ function loadItems(n) {
 }
 
 var intersectionObserver = new IntersectionObserver(entries => {
-  // If intersectionRatio is 0, the sentinel is out of view
-  // and we do not need to do anything.
-  if (entries[0].intersectionRatio <= 0) {
-    return;
-  }
-  loadItems(10);
-  // appendChild will move the existing element, so there is no need to
-  // remove it first.
-  scroller.appendChild(sentinel);
-  loadItems(5);
-  ChromeSamples.setStatus('Loaded up to item ' + counter);
+  // If the browser is busy while scrolling happens, multiple entries can
+  // accumulate between invocations of this callback, so we loop over
+  // them.
+  entries.forEach(entry => {
+    // If intersectionRatio is 0, the sentinel is out of view
+    // and we do not need to do anything.
+    if (entry.intersectionRatio > 0) {
+      loadItems(10);
+      // appendChild will move the existing element, so there is no need to
+      // remove it first.
+      scroller.appendChild(sentinel);
+      loadItems(5);
+      ChromeSamples.setStatus('Loaded up to item ' + counter);
+
+      // We have already loaded more content, no need to look at any more
+      // entries.
+      return;
+    }
+  });
 });
 intersectionObserver.observe(sentinel);
--- a/intersectionobserver/demo.js
+++ b/intersectionobserver/demo.js
@@ -14,23 +14,17 @@ function loadItems(n) {
 
 var intersectionObserver = new IntersectionObserver(entries => {
   // If the browser is busy while scrolling happens, multiple entries can
-  // accumulate between invocations of this callback, so we loop over
+  // accumulate between invocations of this callback. As long as any one
+  // of the notifications reports the sentinel within the scrolling viewport,
+  // we add more content.
   // them.
-  entries.forEach(entry => {
-    // If intersectionRatio is 0, the sentinel is out of view
-    // and we do not need to do anything.
-    if (entry.intersectionRatio > 0) {
-      loadItems(10);
-      // appendChild will move the existing element, so there is no need to
-      // remove it first.
-      scroller.appendChild(sentinel);
-      loadItems(5);
-      ChromeSamples.setStatus('Loaded up to item ' + counter);
-
-      // We have already loaded more content, no need to look at any more
-      // entries.
-      return;
-    }
-  });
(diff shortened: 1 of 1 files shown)


intersectionobserver/demo.js
```diff
@@ -13,16 +13,24 @@ function loadItems(n) {
...
+      loadItems(10);
+      // appendChild will move the existing element, so there is no need to
+      // remove it first.
+      scroller.appendChild(sentinel);
+      loadItems(5);
+      ChromeSamples.setStatus('Loaded up to item ' + counter);
+
+      // We have already loaded more content, no need to look at any more
+      // entries.
+      return;
```
surma: This return only returns from the callback, which ends here anyway. The comment implies that the intent is to stop the `forEach` loop? Not quite sure

intersectionobserver/demo.js
```diff
@@ -13,16 +13,24 @@ function loadItems(n) {
...
-  }
-  loadItems(10);
-  // appendChild will move the existing element, so there is no need to
-  // remove it first.
-  scroller.appendChild(sentinel);
-  loadItems(5);
-  ChromeSamples.setStatus('Loaded up to item ' + counter);
+  // If the browser is busy while scrolling happens, multiple entries can
+  // accumulate between invocations of this callback, so we loop over
+  // them.
```
surma: Oh really? If the browser is busy, can layout really happen? I think layout is required to get the sentinel back below the threshold so that it can then cross the threshold again, doesn’t it?

surma (changes requested):

intersectionobserver/demo.js
szager-chromium: The task responsible for delivering notifications can be de-prioritized, which could cause notifications to accumulate.

intersectionobserver/demo.js
szager-chromium: Right, this 'return' statement ends the forEach loop.

intersectionobserver/demo.js
szager-chromium: You can have a sequence like:

load handler
animation frame
load handler
animation frame
load handler
animation frame
IntersectionObserver notification delivery

intersectionobserver/demo.js
surma: I feel like I’m missing something obvious. How is this `return` statement ending the `forEach`? You can’t break out of a `forEach`.

intersectionobserver/demo.js
surma: Of course, thank you.

intersectionobserver/demo.js
fabb: Exactly. You would need to use `for...of` or a regular `for` loop to being able to break out. The problem with `forEach` is that the closure is executed for every element, and a `return` in the end of that closure has the same effect as having no such `return`.

Alternatively, you could also use `if(entries.some(entry => entry.intersectionRatio > 0)) {...}`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

> There is no way to stop or break a forEach() loop other than by throwing an exception. If you need such behavior, the forEach() method is the wrong tool.

surma: +1 to `some()`

szager-chromium: *facepalm*

This is why I should never write javascript.

I switched to 'some', thanks for the suggestion.

fabb (approved):

fabb:
@szager-chromium the example works better now, but still has the bug sometimes. I think it is related on the sentinel being too high up in the item list, and when scrolling quickly enough, the browser might jump over the sentinel when scrolling without it ever intersecting. See this screenshot for strong evidence for this case:

Screenshot 2020-03-14 at 09 23 44

I think to resolve this, either the sentinel needs to be placed nearer to the end of the list, so it definitely is visible when the end of the list is reached, or we need a second sentinel at the end of the list with more involved logic.

The first solution is easily done in the example by changing the `loadItems(5)` to `loadItems(3)`. In a real world application where the loading would be asynchronous, this would load the external resource a bit later, and could result in a bit more user-perceived loading time. I think fixing this would make the second solution necessary with multiple sentinel elements.

Btw. I have a similar problem in my website, but not with infinite scrolling, but with showing a sticky header after scrolling past some element. I have implemented it with an observed element,  and in the callback checking if the element is becoming invisible *above* the view port by checking for `boundingClientRect.top < rootBounds.top`. The problem is that when this element initially is not visible, and then the browser scrolls down without animation below the element (i.e. when an url `#fragment` is used), and the observed element is not visible again, the intersection observer does not fire. The only way to resolve this I found, without resorting to a scroll observer, is by using a second intersection observer, and check if both the first element is **not** visible and the whole area below **is** visible.

szager-chromium:
@fabb Yes, it's definitely possible to miss a notification if the sentinel moves from below the viewport to above the viewport during a single frame. This is more likely to happen when the browser is busy and scrolling happens on the compositor.

There are a couple of possible mitigations:

- Use the rootMargin argument to the IntersectionObserver to effective observe a large area, making it less likely that the browsers misses the sentinel going from below to above the viewport.

- Use more than one sentinel. Ideally, the number and placement of sentinels should be determined by the number of list items that fit inside the viewport.

fabb:
The first suggestion sounds very good, the sentinel could always put at the end of the list then.

PR #684, opened 2020-05-03
Fix ReferenceError in Web NFC example

1valdis:
Currently, if there's an error while scanning for NFC, it triggers error handler which has no `error` argument. But inside error handler, `error.message` is accessed, which triggers ReferenceError (no `message` property on undefined).

### DIFF
--- a/web-nfc/index.js
+++ b/web-nfc/index.js
@@ -6,8 +6,8 @@ scanButton.addEventListener("click", async () => {
     await reader.scan();
     log("> Scan started");
 
-    reader.addEventListener("error", (error) => {
-      log(`Argh! ${error.message}`);
+    reader.addEventListener("error", (event) => {
+      log(`Argh! ${event.message}`);
     });
 
     reader.addEventListener("reading", ({ message, serialNumber }) => {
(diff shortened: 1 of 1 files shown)


1valdis:
@googlebot I signed it!

web-nfc/index.js
```diff
@@ -6,7 +6,7 @@ scanButton.addEventListener("click", async () => {
     await reader.scan();
     log("> Scan started");
 
-    reader.addEventListener("error", () => {
+    reader.addEventListener("error", (error) => {
```
beaufortfrancois: Thanks for catching this typo!
I'd suggest though we use `event` instead of `error`

```suggestion
    reader.addEventListener("error", (event) => {
```
And use ```log(`Argh! ${event.message}`);```

beaufortfrancois (changes requested):
LGTM with nit

web-nfc/index.js
1valdis: @beaufortfrancois Done.

beaufortfrancois (approved):
Thank you @1valdis!

PR #707, opened 2020-10-19
Include await before writer.close()

guest271314:
`WritableStreamDefaultWriter.close()` returns a `Promise` 
https://streams.spec.whatwg.org/#writable-stream-close

### DIFF
--- a/quictransport/client.js
+++ b/quictransport/client.js
@@ -86,7 +86,7 @@ async function sendData() {
         let stream = await transport.createSendStream();
         let writer = stream.writable.getWriter();
         await writer.write(data);
-        writer.close();
+        await writer.close();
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
It's not because it returns a promise that we necessarily want to wait until it's closed before adding logs.
I'll let @jeffposnick weigh in.

guest271314:
Yes, I gathered that from the code. It might not be immediately obvious to users that adjust the code that `write()` returns a `Promise`. If that is the case  `addToEventLog()` can be called on the line following `let stream = await transport.createSendStream()`.

jeffposnick:
`await`ing sounds fine to me.

Atom1985 (commented):
quictransport/client.js

Atom1985 (commented):
quictransport/client.js

PR #719, opened 2021-04-29
Add Manufacturer Data Filter Web Bluetooth sample

beaufortfrancois:
This sample adds a Web Bluetooth sample for the upcoming Manufacturer Data filter. See https://github.com/WebBluetoothCG/web-bluetooth/pull/545

Live preview: 
- https://beaufortfrancois.github.io/samples/web-bluetooth/manufacturer-data-filter-async-await.html
- https://beaufortfrancois.github.io/samples/web-bluetooth/manufacturer-data-filter.html
- https://beaufortfrancois.github.io/samples/web-bluetooth/index.html

@reillyeon Can you review?

image


web-bluetooth/manufacturer-data-filter-async-await.js
```diff
@@ -0,0 +1,49 @@
+async function onButtonClick() {
+  let manufacturerData = {};
```
reillyeon, line 2: nit: Const, here and for most other variables in the example.

web-bluetooth/manufacturer-data-filter.js
```diff
@@ -0,0 +1,49 @@
+function onButtonClick() {
+  let manufacturerData = {};
```
reillyeon, line 2: Same here.

reillyeon (approved):

web-bluetooth/manufacturer-data-filter-async-await.js
beaufortfrancois: I've replaced `let` with `const` when appropriate in other parts of this example. 
This one was not replaced as I wanted to convey this variable would be updated later.

web-bluetooth/manufacturer-data-filter.js
beaufortfrancois: Ditto.

ttloves2 (commented):
AlphaSTAR

PR #747, opened 2021-10-29
A small update that allows text to be pasted manually into the textbo…

maquarius:
…x and fixes the error that happend when pasting manually.

If it is intended to now allow manual pasting then the e.preventDefault() needs to come back.
"navigator.clipboard.getText()" doesn't work nontheless


maquarius:
@googlebot  I signed it!

maquarius:
@googlebot I signed it!

maquarius:
wrong author email. So close this one and make again

PR #748, opened 2021-10-29
A small update that allows text to be pasted manually into the textbo…

maquarius:
…x and fixes the error that happend when pasting manually.

If the manual pasting is prevented intentionally then e.preventDefault(); needs to be reverted.
navigator.clipboard.getText() doesn't work nontheless


maquarius:
@googlebot I fixed it.

maquarius:
@developit Tagging you since you worked on the project.

developit:
LGTM!

developit (approved):

PR #750, opened 2021-11-03
perf: Unify XHR requests to fetch.

linbuxiao:
During my practice of how to use `service-worker`, I found that the `XHR` in this part seemed abrupt since all the other parts used `fetch`.

Also I modified the html - because the `input` tag was not being used properly. According to the `html` standard, the `input` tag should not have a closing section. This error will cause an error in the template compiler similar to `vite`.


linbuxiao:
@googlebot I signed it!

PR #751, opened 2021-11-05
[WebTransport] Add the protocol version to response

yutakahirano:
...to be aligned with
https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-02.

### DIFF
--- a/webtransport/webtransport_server.py
+++ b/webtransport/webtransport_server.py
@@ -211,6 +211,8 @@ class WebTransportProtocol(QuicConnectionProtocol):
                        status_code: int,
                        end_stream=False) -> None:
         headers = [(b":status", str(status_code).encode())]
+        if status_code == 200:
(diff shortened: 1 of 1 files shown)


yutakahirano:
@jeffposnick @bashi PTAL

bashi (approved):

jeffposnick:
Thanks!

aboba:
The problem seems to have re-emerged in Canary M98.0.4696.0. 

This can be reproduced on:
https://webrtc.internaut.com/wt2/ which is running the python server.  

It runs on Chrome M95, but not on Canary M98.

bashi:
Thank you for the report. Yeah we also noticed that WPTs started failing recently. We are investigating the cause. https://bugs.chromium.org/p/chromium/issues/detail?id=1268607

PR #753, opened 2021-11-05
Fix receiver demo

markafoltz:
Adds more explanation about how cast: URLs are handled by Chromecast.

See https://bugs.chromium.org/p/chromium/issues/detail?id=1267048#c9 for more background.


markafoltz:
CC @takumif

beaufortfrancois (approved):

beaufortfrancois:
Thank you @mfoltzgoogle!

PR #755, opened 2021-11-14
More text and change hyphenated word to fix demo

hugovk:
# now

Even at minimum width, the Chrome (95, desktop) window is too wide to trigger the text to break and demo the hyphens:

image

https://googlechrome.github.io/samples/css-hyphens/index.html

# PR

Double the text to allow it to break, and also move the soft hyphen into "ipsum" so it may break too:

image

https://hugovk.github.io/samples/css-hyphens/

(Just so happens this now demonstrates a Chromium tofu bug, reported here: https://bugs.chromium.org/p/chromium/issues/detail?id=1270110)


hugovk:
Closing this year old PR but happy to re-open if there's interest.

PR #764, opened 2022-02-03
[webtransport] remove unnecessary code

jlaine:
aioquic 0.9.19 already contains all the necessary settings for
WebTransport, so there is no reason to subclass H3Connection.

### DIFF
--- a/webtransport/webtransport_server.py
+++ b/webtransport/webtransport_server.py
@@ -33,7 +33,7 @@ This will output "13" (the length of "Hello, world!") into the console.
 
 # ---- Dependencies ----
 #
-# This server only depends on Python standard library and aioquic 0.9.15 or
+# This server only depends on Python standard library and aioquic 0.9.19 or
 # later. See https://github.com/aiortc/aioquic for instructions on how to
 # install aioquic.
 #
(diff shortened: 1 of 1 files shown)


jlaine:
@yutakahirano @bashi I suspect the WPT code can also be simplified in this manner

jeffposnick:
If @yutakahirano or @bashi can give their 👍  from a technical point of view, I'm happy to get this merged.

bashi:
LGTM, nice cleanup! I haven't tested locally though. I'll try to update wpt server when I have time (I'm a bit swamped these days)

yutakahirano:
Thank you for the change!

bashi:
I realized that we can't update aioquic to 0.9.19 in WPT tools for now since the latest aioquic requires Python 3.7+ but WPT tools need to support Python 3.6.

jlaine:
> I realized that we can't update aioquic to 0.9.19 in WPT tools for now since the latest aioquic requires Python 3.7+ but WPT tools need to support Python 3.6.

That's unfortunate, Python 3.6 reached end of life at the end of 2021, it's no longer maintained. That's the reason I dropped support for Python 3.6 in aioquic and all my projects.

aboba:
Question: The sample WebTransport client (https://github.com/googlechrome/samples/tree/gh-pages/webtransport) no longer seems to work with the latest WebTransport server.  Has something changed?

bashi:
Looks like it's working on my local environment. Could you force-reload then try again? Are there any errors in the DevTools console?

PR #775, opened 2022-06-08
Update index.js

rajababubula:
Updated code to check the record type and parse if text or URL

### DIFF
--- a/web-nfc/index.js
+++ b/web-nfc/index.js
@@ -1,19 +1,15 @@
   scanButton.addEventListener("click", async () => {
   log("User clicked scan button");
-
   try {
     const ndef = new NDEFReader();
     await ndef.scan();
     log("> Scan started");
-
     ndef.addEventListener("readingerror", () => {
       log("Argh! Cannot read data from the NFC tag. Try another one?");
     });
-
     ndef.addEventListener("reading", ({ message, serialNumber }) => {
       log(`> Serial Number: ${serialNumber}`);
-      log(`> Records: (${message.records.length})`);
-      
+      log(`> Records: (${message.records.length})`);    
       for (const record of message.records) {
           log(`\nRecord id: ${record.id}`);
           log(`Record type: ${record.recordType}`);
--- a/web-nfc/index.js
+++ b/web-nfc/index.js
@@ -1,4 +1,4 @@
-scanButton.addEventListener("click", async () => {
+  scanButton.addEventListener("click", async () => {
   log("User clicked scan button");
 
   try {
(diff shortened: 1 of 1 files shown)


web-nfc/index.js
```diff
@@ -13,8 +13,32 @@ scanButton.addEventListener("click", async () => {
     ndef.addEventListener("reading", ({ message, serialNumber }) => {
       log(`> Serial Number: ${serialNumber}`);
       log(`> Records: (${message.records.length})`);
-    });
-  } catch (error) {
+      
+      if(message.records.length>0){
```
beaufortfrancois: This line is not needed.

web-nfc/index.js
```diff
@@ -13,8 +13,32 @@ scanButton.addEventListener("click", async () => {
...
+      for (const record of message.records) {
+          log("Record type:  " + record.recordType);
```
beaufortfrancois: Please use ``log(`Record type: ${record.recordType}`);`` there and below

web-nfc/index.js
```diff
@@ -13,8 +13,32 @@ scanButton.addEventListener("click", async () => {
...
+            case "text":
+              {
+                const textDecoder = new TextDecoder(record.encoding);
+                log(`Text: ${textDecoder.decode(record.data)} (${record.lang})`);
+              }
+              break;
+            case "url":
+              {
+                const textDecoder = new TextDecoder();
+                console.log(`URL: ${textDecoder.decode(record.data)}`);
```
beaufortfrancois: Use `log`, not `console.log`

web-nfc/index.js
```diff
@@ -1,4 +1,4 @@
-scanButton.addEventListener("click", async () => {
+  scanButton.addEventListener("click", async () => {
```
beaufortfrancois, line 1: I don't think this is needed right?

web-nfc/index.js
```diff
@@ -13,8 +13,32 @@ scanButton.addEventListener("click", async () => {
...
+                const textDecoder = new TextDecoder();
+                console.log(`URL: ${textDecoder.decode(record.data)}`);
+              }
+              break;
+            default:
+              // TODO: Handle other records with record data.
+            }
+          }
+         }
+       });
```
beaufortfrancois, line 35: Shall we have a spacer to make it clear there are multiple records?

beaufortfrancois (commented):
Thanks for the PR. Please have a look at my comments.

web-nfc/index.js
rajababubula: I did not modify this line..

rajababubula: This line is needed as it is the listener..

web-nfc/index.js
rajababubula: Agree.. we can remove it.

web-nfc/index.js
rajababubula: Will follow the proposal. thanks

web-nfc/index.js
rajababubula: Sorry this was a mistake..

web-nfc/index.js
rajababubula: This is a good idea.  I want to further parse other types of tags as well.  is this fine?

web-nfc/index.js
beaufortfrancois: I meant the 2 spaces added. Sorry for the confusion.

web-nfc/index.js
beaufortfrancois: I'm not sure I'd parse more types in a sample like this.

beaufortfrancois (commented):
I got an error `Uncaught SyntaxError: missing ) after argument list`.
Please try your PR locally.

805983439:
> Updated code to check the record type and parse if text or URL

Necesito código universal

805983439 (approved):

PR #814, opened 2023-10-03
quick fix - servie-worker.js

yakshjethva:
where the then() method expects to receive a callback function, but you are calling a function from there.

.then(() => self.skipWaiting())


yakshjethva:
where the then() method expects to receive a callback function, but you are calling a function from there. 

Issue : Facing issue in safari & firefox browser.

Proposed Solution: .then(() => self.skipWaiting())

PR #852, opened 2024-09-26
Fix background-blur.js

markafoltz:
This makes the following fixes:

1. Use a normal constraint instead of an advanced constraint to change the blur setting.
2. The result of `applyConstraints()` can only be observed via a `configurationchange` event.
3. Use async/await for `applyConstraints()`

I did not test this yet on a PC with the constraint implemented though.


markafoltz:
This is ready for review @riju @beaufortfrancois

beaufortfrancois:
I've tried this code on macOS by running Chrome with `--enable-experimental-web-platform-features --use-fake-device-for-media-stream` and it works great! Thank you @markafoltz!

PR #868, opened 2024-10-14
Fix readme.md and index.html for deferred-banner sample.

markafoltz:
- Fix link to chromestatus
- Correct text about what happens with the flag
- Correct text about what happens with the button
- Fix URL to demo

### DIFF
--- a/app-install-banner/deferred-banner/index.html
+++ b/app-install-banner/deferred-banner/index.html
@@ -32,23 +32,16 @@ limitations under the License.
     </p>
 
     <ul>
-      <li>The app uses a <a href="https://developers.google.com/web/fundamentals/getting-started/primers/service-workers">service worker</a></li>
       <li>The site is using HTTPS</li>
       <li>The app has a manifest declared</li>
       <li>The manifest has a <code>short_name</code>, 144 pixel icon and a type of 'image/png'</li>
     </ul>
     
-    <p>Frequently developers want be able to control when the Add to homescreen prompt is shown to users.
-       This API and example shows you how you can defer the prompt until a later time, in this case when the 
-       user clicks the button below.  If the button below is disabled, Chrome has chosen now is not a good time to 
-       show the prompt.</p>
+    <p>Frequently developers want be able to control when the install flow is shown to users.
+       This API and example shows you how you can defer the flow until a later time, in this
+       case when the user clicks the button below.</p>
 
-    <p>If the button is disabled, come back in a few minutes and the button will be enabled.</p>
-
-    <p>For testing, we encourage you to force the button to be enabled by setting the
-    chrome://flags/#bypass-app-banner-engagement-checks flag.</p>
-       
-    <button id="btnSave" disabled>Click this to show the prompt</button>
+    <button id="btnSave" disabled>Click this to install the app</button>
     
     <p>Below you will see the output of the example.</p>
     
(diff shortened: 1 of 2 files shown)


app-install-banner/deferred-banner/readme.md
```diff
@@ -1,13 +1,13 @@
...
+See https://googlechrome.github.io/samples/app-install-banner/deferred-banner/index.html for a live demo.
 
 
 #### Directions:
 
-* Install Chrome, or another browser that supports Service Workers and manifest
-* Visit the site over two separate days.
-* Or, more usefully, enable chrome://flags/#bypass-app-banner-engagement-checks
+* Install Chrome, or another browser that supports Service Workers and manifest.
+* Wait a few minutes.
```
markafoltz: @beaufortfrancois I did not have to wait for the button to be enabled on my local server (ChromeOS)

app-install-banner/deferred-banner/index.html
```diff
@@ -42,16 +42,18 @@ <h1>Web App Install Banner: Defer prompt of banner sample</h1>
        This API and example shows you how you can defer the prompt until a later time, in this case when the 
        user clicks the button below.  If the button below is disabled, Chrome has chosen now is not a good time to 
        show the prompt.</p>
+
+    <p>If the button is disabled, come back in a few minutes and the button will be enabled.</p>
+
+    <p>For testing, we encourage you to force the button to be enabled by setting the
```
beaufortfrancois: We may want to wait a bit before updating this sample as @dmurph is removing the engagement requirement for the `beforeinstallprompt` event in Chrome 131

app-install-banner/deferred-banner/readme.md
beaufortfrancois: I wonder if it's still true... @dmurph may know

app-install-banner/deferred-banner/readme.md
```diff
@@ -1,13 +1,13 @@
...
+* Or, enable chrome://flags/#bypass-app-banner-engagement-checks to enable the button right away.
 
-On the second vist, you will see a message on the page that says the user either installed the app or dismissed the banner.
\ No newline at end of file
+Once the button is enabled, Chrome will display the Add to Home Screen prompt when you click it.
```
beaufortfrancois: "Add to Home Screen" is for mobile. On Desktop, it's called "Install App". Maybe use "Chrome will trigger the install flow when you click it"?

app-install-banner/deferred-banner/readme.md
dmurph: You shouldn't need to wait, it should just promote right away if you have the other requirements (display set, start_url set, icon that satisfies the requirement, etc)

dmurph: Also - ServiceWorkers are no longer required in Chrome (maybe also Chromium in general, IDK about Edge) for promotion.

app-install-banner/deferred-banner/readme.md
markafoltz: Updated these files to refer to "install flow"

app-install-banner/deferred-banner/readme.md
markafoltz: Removed references to service workers, the flag, and waiting for the button to be activated

beaufortfrancois (approved):

PR #886, opened 2025-08-27
Import one more PWA related test site from glitch

mkruisselbrink:
While this is started out as a fork of one of the other test sites, this one has some features the original doesn't and we're using it in some manual tests, so would be good to include it as well.


mkruisselbrink:
@tomayac thanks for adding all the other pwa test sites, here is one more that was missing from the original set

pwa-testing/lofty-transparent-donut/index.html
```diff
@@ -0,0 +1,142 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="stylesheet" href="./style.css" />
+    <meta name="theme-color" content="#3f51b5" />
+    <link rel="manifest" href="./manifest.json" />
+    <script>
+      // Redirect to HTTPS if HTTP is requested.
+      if (window.location.protocol === "http:") {
+        window.location.href = "https:" + window.location.href.substring(5);
+      }
+    </script>
```
tomayac:
```suggestion
```

tomayac: This makes testing on `localhost` annoying, and is no longer needed, as GitHub pages is always `https:`.

pwa-testing/lofty-transparent-donut/manifest.json
```diff
@@ -0,0 +1,48 @@
...
+      "type": "image/png"
+    },
+    {
+      "src": "fugu-monochrome-weird.png",
+      "sizes": "512x512",
+      "type": "image/png",
+      "purpose": "monochrome"
+    }
+  ],
+  "scope": "/",
```
tomayac:
```suggestion
  "scope": "./",
```

pwa-testing/lofty-transparent-donut/manifest.json
```diff
@@ -0,0 +1,48 @@
...
+  "start_url": "/index.html",
```
tomayac:
```suggestion
  "start_url": "./index.html",
```

tomayac (approved):
LGTM, thank you.

I'll just apply the suggestions myself and merge.

PR #887, opened 2025-09-15
Fix index.html to point to a manifest for pwa-icon-size-tester

Dp-Goog:
Without this, https://googlechrome.github.io/samples/pwa-testing/pwa-icon-size-tester/ is not installable. The service worker fetching code is responsible for returning a valid manifest.

Verified on localhost.

WorkingInstallable


tomayac (approved):
LGTM, thank you!

PR #892, opened 2025-10-01
Fix manifest id for a couple of PWA samples.

mkruisselbrink:
Since the manifest id field is resolved relative to the origin of the start URL, any sample that sets it to "." or "./" gets the same effective id. While explicitly setting the manifest id is generally considered a good practice, for these samples not setting the id and relying on the manifest id being derived from the start url is simpler.


tomayac (approved):
LGTM

PR #893, opened 2025-10-01
Fix extended scope PWA test site

mkruisselbrink:
It's probably not the best long-term solution to have the extended scope site this links to be part of my personal github repo, the requirement of it containing a top-level .well-known directory makes hosting the second site within this repo tricky. With these changes at least for now this sample works again.


tomayac (approved):
LGTM

PR #897, opened 2025-10-06
fix localSession variable name

mineralsfree:


beaufortfrancois:
LGTM. I'll let @tomayac review as well.

tomayac (approved):
LGTM, thanks for the catch!

tomayac:
Also fixed this in the [accompanying article](https://developer.chrome.com/docs/ai/inform-users-of-model-download#:~:text=false%3B%0Alet-,session,-%3D%20null%3B) (may take some time to be pushed live, still in review).

PR #921, opened 2026-01-26
Update ServiceWorker Static Routing site to use spec align names

quasi-mod:
This PR updates the ServiceWorker static routing API demo site to use spec aligned field names instead of the names which were used during the Origin Trial.


quasi-mod:
@sisidovski , could you take a look at the PR? Thanks!

sisidovski (approved):

tomayac (approved):
LGTM, thank you

PR #923, opened 2026-01-29
Add entry for predictable app updating test site on https://googlechrome.github.io/samples/pwa-testing/

Dp-Goog:
I missed this while landing the predictable app updating test site here, forgot to update the index page with the description. Better late than never!


tomayac (approved):
LGTM, good catch. I didn't think about it either.

PR #925, opened 2026-02-02
Adding test site for manifest localization

kyerebo:


pwa-testing/manifest-localization/index.html
```diff
@@ -0,0 +1,248 @@
+<!DOCTYPE html>
```
Dp-Goog: Let's follow the [naming pattern](https://github.com/GoogleChrome/samples/tree/gh-pages/SAMPLE_STARTING_POINT) followed here? Maybe we can make the folder name to be `manifest-localization-test`?

This is kind of how we did it in the past, see [the test site for predictable app updating](https://github.com/GoogleChrome/samples/tree/gh-pages/pwa-testing/predictable-app-update) here.

pwa-testing/manifest-localization/index.html
```diff
@@ -0,0 +1,248 @@
...
+        </table>
+      </div>
+
+
+    </div>
+  </div>
+
+  <script>
+    // Register service worker
+    if ('serviceWorker' in navigator) {
```
Dp-Goog: Can we move the script to a separate `main.js`?

HTML styleguide for Google suggests separating structure, presentation and behavior [1].

[1] https://google.github.io/styleguide/htmlcssguide.html#Separation_of_Concerns

Dp-Goog (changes requested):
Requested some minor changes, but otherwise looks good.

@tomayac since you have more experience in this repo, can you take a look please? Alex is from Microsoft and is helping us implement [manifest localization](https://www.w3.org/TR/appmanifest/#x_localized-members).

Dp-Goog:
Also, don't forget to update the `index.html` at the top level folder of `pwa-testing` [1] by adding an entry to this metadata here, so that it shows up on https://googlechrome.github.io/samples/pwa-testing/ with the site name and some details about it. I had forgotten to do this earlier haha!

[1] https://github.com/GoogleChrome/samples/blob/gh-pages/pwa-testing/index.html#L92

pwa-testing/manifest-localization/index.html
kyerebo: I completely missed the starting point doc, thanks for bringing it to my attention.
I have a couple questions about it-

Is the 3-part folder name a standard? I'm seeing other test sites that break that convention so I was wondering what the motivation was behind the `manifest-localization-test` suggestion.

The index.html has the following code block:
```
---
feature_name: REPLACE_ME
chrome_version: REPLACE_ME
feature_id: REPLACE_ME
local_css_files: ['optional_relative_path.css', 'anotheroptional_relative_path.css']
local_js_files: ['optional_relative_path.js', 'anotheroptional_relative_path.js']
additional_head_content: <meta name="optional" value="optional">
---
```
but I don't see that in the `pending-app-update` test site or other test sites, is it required?
Thanks in advance for the clarification!

Dp-Goog: I think for the test site itself, the code in `index.html` or the manifest/JS files are fine to not have to follow the convention? I didn't do that in the past.

As for the folder name, I think because a lot of these sites have been migrated over from Glitch, and Glitch automatically gave them a 3 word name, it was easier to follow that convention here. But I do see a lot of examples that don't follow that convention, so I don't have a strong opinion here.

Either way, I'll defer to Thomas here for guidance, since he has worked in this repo more and might have more information.

Dp-Goog:
This LGTM, thanks! Feel free to resolve the comments above.

Dp-Goog (approved):

tomayac (approved):
LGTM with important nit:

- The paths all assume root `/`, but they will end up being relative to `https://googlechrome.github.io/samples/pwa-testing/manifest-localization-test/`.

PR #930, opened 2026-02-09
Caching icons with the service worker for offline use case

kyerebo:


Dp-Goog (approved):
Thanks for adding the icons caching code, let's see what happens!

